PR #14, opened 2019-11-24
Support for nested and circular geometry types and more

bjornharrtell:
This is an attempt to support heterogeneous collections fx. GeometryCollection and curved geometries fx. CompoundCurve.

It also adds a property value type Binary and aligns to use the same geometry type enumeration as found in WKB.

It also optionally allows per feature schema.

This is a breaking change so spec is bumped to version 2.


src/ts/header_generated.ts
```diff
@@ -5,12 +5,21 @@ import { flatbuffers } from 'flatbuffers'
...
+  Polygon= 3,
+  MultiPoint= 4,
+  MultiLineString= 5,
+  MultiPolygon= 6,
+  GeometryCollection= 7,
+  CircularString= 8,
+  CompoundCurve= 9,
+  CurvePolygon= 10,
+  MultiCurve= 11,
+  MultiSurface= 12,
```
rouault: as a comment, perhaps add that 13 and 14 are reserved values for the abstract Curve and Surface types ? (so that they don't get accidentally reused for other purposes)

src/ts/header_generated.ts
```diff
@@ -5,12 +5,21 @@ import { flatbuffers } from 'flatbuffers'
...
+  Point= 1,
   LineString= 2,
-  MultiLineString= 3,
-  Polygon= 4,
-  MultiPolygon= 5
+  Polygon= 3,
+  MultiPoint= 4,
+  MultiLineString= 5,
+  MultiPolygon= 6,
+  GeometryCollection= 7,
```
rouault: as GeometryCollection means a multi-geometry of unknown simple types, it could be consistent to also have a 0 = Unknown

bjornharrtell: Agreed, makes sense to have Unknown in several aspects.

src/ts/header_generated.ts
bjornharrtell: Agreed, fixed.

PR #20, opened 2019-12-22
RFC: Use JTS PackedCoordinateSequence (Do not merge)

jagill:
When deserializing flatgeobuff into a Java JTS geometry, we pack apart
the nice `byte[]` array into a `Coordinate[]` array, which allocates
many objects on the heap and scatters the memory losing data locality.
JTS provides a `PackedCoordinateSequence` type, which stores the
coordinates in a packed `double[]`.  This is much more friendly to the
garbage collector, and also has a smaller memory footprint (no object
overhead) and is generally faster (data locality results in fewer L1/2/3
cache misses).

This is a POC that uses the `PackedCoordinateSequence` instead of the
default `Coordinate[]` form.  It seems sad to scatter the bytes after
so much work for zero-copy.  It'd be nice if the `byte[] -> double[]`
conversion could also be zero-copy, but the only method I know to do
that involves `unsafe`.

NB: This does not handle Z/M coordinates at all.  This is more of a
conversation starter to see if this is a direction worth pursuing.


jagill:
cc @dr-jts

bjornharrtell:
This definitely seems to be a good idea.

nicolas-f:
Current implementation use  CoordinateSequenceFilter, maybe this PR could be closed ?

bjornharrtell:
@nicolas-f I do not see the relation to CoordinateSequenceFilter?

nicolas-f:
> @nicolas-f I do not see the relation to CoordinateSequenceFilter?

You fetch the coordinates using CoordinateSequenceFilter of Jts that iterate over coordinates instead of creating a new array.

nicolas-f:
Oh sorry I was thinking it was an old code when converting from JTS. I should have looked into that when I was working on z m ordinates.

PR #22, opened 2020-01-05
WIP: Use PackedCoordinateSequence

bjornharrtell:
Adaptation of https://github.com/bjornharrtell/flatgeobuf/pull/20.

Still working on Z/M support. It will not be optimal as PackedCoordinateSequence uses a single array whereas flatgeobuf has separate arrays for Z and M respectively.

NOTE: In the ecosystem we see this difference in memory layout of coordinate data and there seem to be no consesus. Fx. JTS and OpenLayers have a single array with dimension as stride where as GDAL and QGIS keep the additional dimensions as separate arrays.

### DIFF
--- a/leaflet/index.html
+++ b/leaflet/index.html
@@ -1,0 +1,15 @@
+<html>
+<head>
+    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.6.0/dist/leaflet.css" />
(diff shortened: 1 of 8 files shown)


bjornharrtell:
One thought would be to implement a variant of PackedCoordinateSequence with the separate array memory model.

bjornharrtell:
Closing due to inactivity.

PR #46, opened 2020-02-29
cpp: packedrtree: add header and resync with GDAL

rouault:
- Add logic for MSB hosts
- Add note with the link to the upstream repository (that is here)


bjornharrtell:
Make sense to have them identical, thanks.

PR #47, opened 2020-02-29
Rust port

pka:
While having a deeper look at flatgeobuf, I started a Rust port. Don't know how much time I can invest, but flatbuf generated code looks already good so far.


bjornharrtell:
Great stuff! It will be fun to take a deeper look at this.

pka:
This is now ready for a first review.

Things I would like to do next:
- Idiomatic reader implementation
- Rust implementation of packed R-tree index
- Library documentation
- More examples/tests
- Publication on crates.io

More plans:
- Async http reader

pka:
Another plan I forgot about:
* Compiling to WASM

@bjornharrtell I've sent you a mail to your gmail adress to find a date for talking about these plans.

bjornharrtell:
Compiling to WASM sounds fun especially to see how it would benchmark against the native TypeScript implementation.

pka:
A first version of zero-copy geometry and feature readers are now implemented. What would be a good benchmark? Maybe a GeoJSON writer implementation to compare with the C++ impementation?

bjornharrtell:
See you are making good progress @pka, nice to see! :)

About Packed Hilbert R-tree vs OMT (vs STR?), I'm aware of that paper but I lack reference implementations of it and real world experience/comparison to alternatives. The only thing I'm certain of is that they are more complex to implement and slower to build the tree.

However, the main reason I went with Packed Hilbert R-tree is because the implementation by @mourner that I've based mine on is the first and only R-tree implementation I've found that I could understand. I find it simple, beautiful and performant. :)

https://github.com/mourner/flatbush/issues/8 is also an interesting observation.

pka:
I'm fine with your desision. What came to my mind was the possible need for other kind of indices, like fulltext or key/value indices. Maybe the FGB format could have a header section allowing multiple indices with a type declaration and offsets to skip them.

bjornharrtell:
I have also thought about other types of indices but came to the conclusion that it adds to much complexity and essentially is a step too far on the way to reimplement a database engine so my thinking is that GeoPackage and/or PostGIS fit that space already.

If you have a dataset where spatial index is secondary to a column index, you should probably split it up. (unless you use PostGIS where you might be better served by a combined geometry+column(s) index)

bjornharrtell:
With that said, the header could perhaps be extended a bit to allow for some custom/extra sections. (also required for https://github.com/bjornharrtell/flatgeobuf/issues/43)

mourner:
> About Packed Hilbert R-tree vs OMT (vs STR?), I'm aware of that paper but I lack reference implementations of it and real world experience/comparison to alternatives.

@bjornharrtell [RBush](https://github.com/mourner/rbush) implements an improved variation of OMT for bulk loading, but it's pretty involved — for static indices, Packed Hilbert is much simpler and more elegant, and fits the case perfectly.

pka:
There are a few criterion benches included now. First measurements on my laptop (Intel Core i7-10510U, SSD):
- Read all features from `countries.fgb` and write them to a GeoJSON file: 28ms
- `time ogr2ogr -f GeoJSON -oo VERIFY_BUFFERS=NO /tmp/countries-ogr.json ../../test/data/countries.fgb`: 80ms

- Read all features from `countries.fgb` and write GeoJSON to /dev/null: 3.8ms
- `time ogr2ogr -f GeoJSON -oo VERIFY_BUFFERS=NO /dev/null ../../test/data/countries.fgb`: 55ms

So pure file writing time seems to be 25ms in both cases. The comparison is not fair because ogr2ogr does more initialization like command line parsing, format detection, etc., but shows that the Rust implementation is not particularly slow, at least.

Recommendations for benchmark improvements are welcome.

pka:
By using `BufWriter` the time for a full countries.fgb read (201KB) + geojson write (253KB) goes down to 4ms which is 20 times faster than ogr2ogr!

After the recent API cleanups, the Rust implementation is now ready to merge.

bjornharrtell:
Great work @pka. I've looked through the code once again and it looks good. I've not had time to play with it but at least I can understand it. :) Will merge.

PR #54, opened 2020-05-09
Rust API and index improvements

pka:
This is a major Rust implementation update:
* Simplify high-level API
* Extract processing API with GeoJSON and SVG implementations into separate crate https://crates.io/crates/geozero
* Implement stream_search for FileReader and HttpReader
* Return Error types instead of unwrap
* Implement remaining property types

Detailed commits are in https://github.com/pka/flatgeobuf/commits/geozero


bjornharrtell:
Looks great, I'll merge.

PR #55, opened 2020-05-10
Add missing files from dd78453

pka:
lost in manual rebase of squashed commit


bjornharrtell:
Hmm not sure if this was due to my squash merge or not, sorry in that case.

Issue #59, opened 2020-05-30
Support Geometry subclasses in Java


jericks:
Currently,  Geometry subclasses in Java do not appear to be supported.

If I create a subclass like this:

`class MyPoint extends Point {
    public MyPoint(CoordinateSequence coordinates, GeometryFactory factory) {
        super(coordinates, factory);
    }
}`

the `GeometryConversions.toGeometryType(MyPoint.class);` will throw a RuntimeException "Unknown geometry type".

I think all we need to do is switch around the order of the classes in the isAssignableFrom method:

Instead of 

`geometryClass.isAssignableFrom(Point.class)`

we need

`Point.class.isAssignableFrom(geometryClass)`

I can work on a PR with tests if this sounds right.

jericks:
Pull request = https://github.com/bjornharrtell/flatgeobuf/pull/60

PR #60, opened 2020-05-31
Add support for Geometry subclasses

jericks:
Solves #59


bjornharrtell:
Makes sense! Thanks again @jericks.

jericks:
Thanks!  When you have time would you consider doing a new release to maven?  I am working on updating the GeoTools DataStore and then adding support in GeoScript.

bjornharrtell:
Sounds like an effort I'd like to support. I've published latest Java package as 3.3.0 to Maven Central, should be available soon.

PR #75, opened 2020-08-28
Additional metadata fields

bjornharrtell:
ref https://github.com/bjornharrtell/flatgeobuf/issues/44, https://github.com/bjornharrtell/flatgeobuf/issues/45 and https://github.com/bjornharrtell/flatgeobuf/issues/73

This is my suggestion to add a set of metadata fields intended to resolve the use cases described in the above referenced issues. @rouault and @palmerj feedback would be much appreciated before I implement this into the reference code and GDAL driver.

NOTE: I believe these additions will be 100% backward compatible with the initial v3 spec.


bjornharrtell:
I've gone ahead and implemented the schema additions in my reference code. Before merging I would still very much like to get your input on the additions and comments in https://github.com/bjornharrtell/flatgeobuf/blob/metadata/src/fbs/header.fbs @rouault and ®palmerj.

palmerj:
Thanks @bjornharrtell. I will look over the next few days. If I want to test in GDAL/OGR is it safe to copy the code from this branch into https://github.com/OSGeo/gdal/tree/master/gdal/ogr/ogrsf_frmts/flatgeobuf/flatbuffers and compile GDAL?

bjornharrtell:
@palmerj should be safe but nothing would be written into the new metadata fields, needs to be wired up in the GDAL specifics. I should be able to do that shortly in a PR to GDAL that I can set as WIP until we are more certain about this metadata.

bjornharrtell:
@palmerj GDAL PR ready, looking forward to your test results.

src/fbs/header.fbs
```diff
@@ -44,6 +44,14 @@ enum ColumnType: ubyte {
...
   name: string (required);      // Column name
   type: ColumnType;             // Column type
+  title: string;                // Column title
+  description: string;          // Column description (intended for free form long text)
+  width: int = -1;              // Column values expected width (-1 = unknown) (currently only used to indicate the number of characters in strings)
+  precision: int = -1;          // Column values expected precision (-1 = unknown) as defined by SQL
+  scale: int = -1;              // Column values expected scale (-1 = unknown) as defined by SQL
+  nullable: bool = true;        // Column values expected nullability
+  unique: bool = false;         // Column values expected uniqueness
+  primary_key: bool = false;    // Indicates this column have been (part) of primary key
```
rouault, line 54:
```suggestion
  primary_key: bool = false;    // Indicates this column has been (part of) a primary key
```
So the intent is that if several columns are flagged primary_key = true, this is a multi-column primary key ?

src/fbs/header.fbs
```diff
@@ -52,6 +60,7 @@ table Crs {
   name: string;                 // Human readable name of this SRS
   description: string;          // Human readable description of this SRS
   wkt: string;                  // Well-known Text Representation of the Spatial Reference System
+  code_string: string;          // Text ID of the Spatial Reference System assigned by the organization (rarely used)
```
rouault, line 63: not clear what that would be. Can you give an example ?

src/fbs/header.fbs
bjornharrtell: Yes

src/fbs/header.fbs
bjornharrtell: This is to resolve https://github.com/bjornharrtell/flatgeobuf/issues/44, I think you know more about it than me? 🙂

rouault: oh !!! my memory is so terrible.  Perhaps:
```suggestion
  code_string: string;          // ID of the Spatial Reference System assigned by the organization, when it is not an integer, and thus cannot be set into code.
```

pka: I did understand it as text part only (e.g. `EPSG`) used together with the numerical code (e.g. `4326`), but in @rouault's example `IGNF:RGF93` there is no numerical part. Maybe it should be documented, which one is used when both the numerical srid and code_string is set.

bjornharrtell: I've attempted to clarify. I've kept the word rare in the sentence to indicate that it's not to be preferred instead of `code` when possible.

PR #84, opened 2020-11-23
Add fuzz target for the Rust crate.

frewsxcv:
To run:

```
cargo install cargo-fuzz
cd src
cd rust
cargo fuzz run read
```


bjornharrtell:
Nice, seems valuable to me - @pka do you agree?

pka:
Sure, it's a great addition. Time to start fuzzing!

PR #89, opened 2020-12-14
Update README.md with developer notes.

evetion:
Also included link to the Julia package.

I'm unsure about the UInt16 offset in the properties of the feature, could you comment on that?

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,13 +27,15 @@ The site [switchfromshapefile.org](http://switchfromshapefile.org) has more in d
 ![layout](doc/layout.svg "FlatGeobuf file layout")
 
 * MB: Magic bytes (0x6667620366676200)
-* H: Header (variable size [flatbuffer](https://github.com/bjornharrtell/flatgeobuf/blob/master/src/fbs/header.fbs))
-* I (optional): Static packed Hilbert R-tree index (static size [custom buffer](https://github.com/bjornharrtell/flatgeobuf/blob/master/src/cpp/packedrtree.h))
-* DATA: Features (variable size [flatbuffer](https://github.com/bjornharrtell/flatgeobuf/blob/master/src/fbs/feature.fbs)s)
+* H: Header (variable size [flatbuffer](https://github.com/bjornharrtell/flatgeobuf/blob/master/src/fbs/header.fbs)), prefixed with the header size as UInt32.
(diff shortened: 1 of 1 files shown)


bjornharrtell:
Does the comment at https://github.com/bjornharrtell/flatgeobuf/blob/4d70a8c5c774fdc39d5fc95a3c36af5803804c87/src/fbs/feature.fbs#L18 help to explain? Properties is a key/value collection where the key is a short (i.e column index). I agree it could/should be better explained..

The uint32 prefixes are flatbuffers standard length prefix so I'm reluctant to reiterate that (I like very short and to the point docs). Although I note it's not clearly documented by upstream (but that's where it belongs).

evetion:
Ah, I missed that. And agreed, upstream documentation is lacking. Given that you like your documentation short and to the point, how do you want to go about developer documentation? If others want to implement flatgeobuf, they need to go through some other implementation and mimic that.

With regards to the properties, the keys, do they mean anything? Columns have no key, so it's only about the order? Can the properties be out of order compared to the columns, or even ordered differently for each feature?

bjornharrtell:
@evetion the key is the index of the column in the header columns array at https://github.com/bjornharrtell/flatgeobuf/blob/4d70a8c5c774fdc39d5fc95a3c36af5803804c87/src/fbs/header.fbs#L75

Properties can be out of order yes and it's intentional because it's how I support null values without taking up any space at all.

A separate sub-page could be done at some point with more in depth information about the spec and implementation with developer focus. Until then the existing reference implementations will have to do.

evetion:
Taking no space at all is a nice thing. I assume this is only allowed if `nullable` is true on a column? And I would expect you get a pattern like `1, 2, 3, 5` where key `4` is missing. Or do your implementations also allow for `1, 5, 3, 2` orders?

I see GDAL doesn't support per feature schemas yet (nor do I), do you expect other drivers will do so?

bjornharrtell:
@evetion I think it makes sense to make no assumptions regarding ordering of the properties but I've not made sure actual implementation can handle it because the produced data is probably always in order. Until proven a problem I think I will not look into it at this time. :)

Per feature schemas support is a bit exotic. Short answer is that I have no need for it myself and as such I'm not planning to implement it myself. It's not the common case for datasets and I'm not sure how much work it would require in the GDAL driver to support it (or even if GDAL has the data model for it). I introduced it in the spec for it to be flexible enough to represent unstructured data in the same way GeoJSON can.

Column metadata like `nullable` is just metadata, not enforced constraints, so driver can/should not make any assumptions based on that metadata and an end user would have to validate data if it's not trusted.

evetion:
Fair enough, thanks for all the answers and feedback. I will at least make sure to support the skipping of properties so the Julia client can handle missing data. I will close this PR for now.

PR #92, opened 2020-12-21
rust client: log network usage

michaelkirk:
This captures some rudimentary usage statistics, and, if a logging backend is set-up, can log lines like this:

```
[2020-12-21T19:27:28Z TRACE flatgeobuf::http_reader] starting: opening http reader, reading header
[2020-12-21T19:27:28Z DEBUG flatgeobuf::http_client] request: #1, bytes: (this_request: 512, ever: 512), Range: bytes=0-511
[2020-12-21T19:27:29Z DEBUG flatgeobuf::http_client] request: #2, bytes: (this_request: 520, ever: 1032), Range: bytes=512-1031
[2020-12-21T19:27:29Z TRACE flatgeobuf::http_reader] completed: opening http reader
[2020-12-21T19:27:29Z TRACE flatgeobuf::http_reader] starting: select_bbox, traversing index
[2020-12-21T19:27:29Z DEBUG flatgeobuf::http_client] request: #3, bytes: (this_request: 1048576, ever: 1049608), Range: bytes=1032-1049607
[2020-12-21T19:27:32Z DEBUG flatgeobuf::http_client] request: #4, bytes: (this_request: 1048576, ever: 2098184), Range: bytes=1982072-3030647
[2020-12-21T19:27:34Z DEBUG flatgeobuf::http_client] request: #5, bytes: (this_request: 1048576, ever: 3146760), Range: bytes=31695952-32744527
[2020-12-21T19:27:37Z DEBUG flatgeobuf::http_client] request: #6, bytes: (this_request: 1048576, ever: 4195336), Range: bytes=33971792-35020367
[2020-12-21T19:27:39Z TRACE flatgeobuf::http_reader] completed: select_bbox
[2020-12-21T19:27:39Z DEBUG flatgeobuf::http_client] request: #7, bytes: (this_request: 1048576, ever: 5243912), Range: bytes=268512880-269561455
```

Is this something you'd be interested in including? It might be helpful to evaluate some of the upcoming changes I'd like to propose around what the client fetches.

PR #93, opened 2020-12-22
rust http client: optimize index fetching 

michaelkirk:
Based on #92, so you'll want to review that one first. Alternatively I can close that one if you'd prefer to just have it all here.

This is based on some of my findings in https://github.com/bjornharrtell/flatgeobuf/issues/90, but excludes any changes which would rely on the not-well-supported `MultiPart Range Requests` for now.

The commits are intended to be separable — each offers independent gains. You might want to review them separately.

The bottom line is:
before: 7 requests totaling 4MB index traversal + 1MB feature fetching
after: 6 requests totaling 512KB index traversal + 1MB feature fetching

This is anecdotal - and depending on where your data lives in the index, you might see more or less requests, but I believe that these changes are rarely worse for small data and often a lot better for large data.

Some context, the file layout is:

`[[Header], [Optional Index], [Features]]`

The header is very small - less than 1kb in my experience.

The optional index, when present, is a tree laid out starting with the root, followed by subsequent layers. Here's the size of each layer assuming branching factor 16:

```
root   = 40*16^0 = 40 bytes
root+1 = 40*16^1 = 640 bytes
root+2 = 40*16^2 = 10KB
root+3 = 40*16^3 = 164KB
root+4 = 40*16^4 = 2.6MB
root+5 = 40*16^5 = 42MB
```

Inlining that into the earlier representation for clarity:
```
[ 
  [Header: <1kb], 
  [[root: 40 bytes], [root+1: 640 bytes], [root+2: 10kb], [164kb], [2.6mb], [42mb], ...], 
  [Features]
]
```

The key insights were: 

1. early layers are very small - try to fetch the first three along with the header

2. The previous approach of fetching extra data to satisfy a minimum request size worked well to eliminate early layers, but now that those early layers are fetched with the header, fetching extra data is unlikely to "randomly" hit useful nodes. Instead we combine nearby requests within a layer, expanding their boundaries up to a reasonable amount, so we can combine the requests.

Notice that there's a "DO NOT MERGE" commit which has some code for logging in tests which I left in for now in case it's helpful to you for the purposes of review.

### DIFF
--- a/src/rust/src/http_client.rs
+++ b/src/rust/src/http_client.rs
@@ -74,7 +74,7 @@ impl BufferedHttpClient {
 
             // Read additional bytes
             let range_begin = max(begin, self.tail());
-            let range_length = max(length - (range_begin - begin), min_req_size);
+            let range_length = max(begin + length - range_begin, min_req_size);
             let bytes = self.http_client.get(range_begin, range_length).await?;
             self.buf.put(bytes);
         }
--- a/src/rust/src/http_client.rs
+++ b/src/rust/src/http_client.rs
@@ -6,6 +6,8 @@ use std::str;
 struct HttpClient {
     client: reqwest::Client,
     url: String,
+    requests_ever_made: usize,
+    bytes_ever_requested: usize,
 }
 
 impl HttpClient {
@@ -13,13 +15,22 @@ impl HttpClient {
         HttpClient {
             client: reqwest::Client::new(),
             url: url.to_string(),
+            requests_ever_made: 0,
+            bytes_ever_requested: 0,
         }
     }
-    async fn get(&self, begin: usize, length: usize) -> Result<Bytes> {
+    async fn get(&mut self, begin: usize, length: usize) -> Result<Bytes> {
+        self.requests_ever_made += 1;
+        self.bytes_ever_requested += length;
+        let range = format!("bytes={}-{}", begin, begin + length - 1);
+        debug!(
+            "request: #{}, bytes: (this_request: {}, ever: {}), Range: {}",
+            self.requests_ever_made, length, self.bytes_ever_requested, range
+        );
         let response = self
             .client
             .get(&self.url)
-            .header("Range", format!("bytes={}-{}", begin, begin + length - 1))
+            .header("Range", range)
             .send()
             .await
             .map_err(|e| GeozeroError::HttpError(e.to_string()))?;
--- a/src/rust/src/http_client.rs
+++ b/src/rust/src/http_client.rs
@@ -74,7 +74,6 @@ impl BufferedHttpClient {
 
             // Read additional bytes
             let range_begin = max(begin, self.tail());
-            debug!("http_client#get begin: {}, tail: {}, length: {}, range_begin: {}, min_req_size: {}", begin, self.tail(), length, range_begin, min_req_size);
             let range_length = max(length - (range_begin - begin), min_req_size);
             let bytes = self.http_client.get(range_begin, range_length).await?;
             self.buf.put(bytes);
--- a/src/rust/src/http_client.rs
+++ b/src/rust/src/http_client.rs
@@ -59,26 +59,33 @@ impl BufferedHttpClient {
             head: 0,
         }
     }
+
     pub async fn get(&mut self, begin: usize, length: usize, min_req_size: usize) -> Result<&[u8]> {
-        let tail = self.head + self.buf.len();
-        if begin + length > tail || begin < self.head {
+        // If we don't already have the data, we download it
+        if begin + length > self.tail() || begin < self.head {
             // Remove bytes before new begin
-            if begin > self.head && begin < tail {
+            if begin > self.head && begin < self.tail() {
                 let _ = self.buf.split_to(begin - self.head);
                 self.head = begin;
-            } else if begin >= tail || begin < self.head {
+            } else if begin >= self.tail() || begin < self.head {
                 self.buf.clear();
                 self.head = begin;
             }
 
             // Read additional bytes
-            let range_begin = max(begin, tail);
-            let range_length = max(begin + length - range_begin, min_req_size);
+            let range_begin = max(begin, self.tail());
+            debug!("http_client#get begin: {}, tail: {}, length: {}, range_begin: {}, min_req_size: {}", begin, self.tail(), length, range_begin, min_req_size);
+            let range_length = max(length - (range_begin - begin), min_req_size);
             let bytes = self.http_client.get(range_begin, range_length).await?;
             self.buf.put(bytes);
         }
+
         let lower = begin - self.head;
         let upper = begin + length - self.head;
         Ok(&self.buf[lower..upper])
+    }
+
+    fn tail(&self) -> usize {
+        self.head + self.buf.len()
     }
 }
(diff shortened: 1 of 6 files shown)


src/rust/src/http_client.rs
```diff
@@ -48,26 +59,32 @@ impl BufferedHttpClient {
...
             let bytes = self.http_client.get(range_begin, range_length).await?;
             self.buf.put(bytes);
         }
+
         let lower = begin - self.head;
         let upper = begin + length - self.head;
         Ok(&self.buf[lower..upper])
     }
+
+    fn tail(&self) -> usize {
```
michaelkirk, line 87: Adding this method fixes an edge case I discovered while messing around with different load windows. See 6129a2ff963822bc6f6f1f8f9326399174575518

src/rust/src/http_reader.rs
```diff
@@ -27,8 +27,33 @@ pub struct HttpFgbReader {
...
+
+        // Because we use a buffered HTTP reader, anything extra we fetch here can
+        // be utilized to skip subsequent fetches.
+        // Immediately following the header is the optional spatial index, we deliberately fetch
+        // a small part of that to skip subsequent requests
+        let prefetch_index_bytes: usize = {
+            // The actual branching factor will be in the header, but since we don't have the header
+            // yet we guess. The consequence of getting this wrong isn't catastrophic, it just means
+            // we may be fetching slightly more than we need or that we make an extra request later.
+            let assumed_branching_factor: usize = 16;
```
michaelkirk, line 41: @pka - you mentioned hoisting some constants to a Config, which I think is a great idea. 

If you'd like I could include that in this PR, or follow up with it later.

Things which might be configurable:

- assumed branching factor

- num_prefetched_layers (not sure if this should be an integer number of layers or a byte count - byte count would probably make more sense to people, but it'd be easy to include 1/2 an index layer, which is unfortunate)

- assumed_header_size?

- combine_request_threshold?

- feature request size

src/rust/src/http_client.rs
```diff
@@ -48,26 +59,32 @@ impl BufferedHttpClient {
...
+            } else if begin >= self.tail() || begin < self.head {
                 self.buf.clear();
                 self.head = begin;
             }
 
             // Read additional bytes
-            let range_begin = max(begin, tail);
-            let range_length = max(begin + length - range_begin, min_req_size);
+            let range_begin = max(begin, self.tail());
+            let range_length = max(length - (range_begin - begin), min_req_size);
```
pka, line 77: This change here is critical. Don't understand it immediately. Will add some comments to my code, to understand it.

src/rust/src/http_reader.rs
```diff
@@ -27,8 +27,33 @@ pub struct HttpFgbReader {
...
+
+            // NOTE: each layer is exponentially larger
+            let prefetched_layers: u32 = 3;
+
+            (0..prefetched_layers).map(|i| {
+                assumed_branching_factor.pow(i) * std::mem::size_of::<NodeItem>()
```
pka, line 47: Should that be the same as level node counts from https://github.com/bjornharrtell/flatgeobuf/blob/master/src/rust/src/packed_r_tree.rs#L300 ?

src/rust/src/packed_r_tree.rs
```diff
@@ -570,8 +582,41 @@ impl PackedRTree {
...
                     });
-                } else {
-                    queue.push(Reverse((node_item.offset as usize, level - 1)));
+                    continue;
+                }
+
+                // spend this many extra bytes if it means we'll eliminate an extra request
+                let combine_request_threshold = 256 * 1024 / size_of::<NodeItem>();
+
+                match (queue.back_mut(), node_item.offset as usize) {
```
pka, line 591: Boah, this is complicated. I trust you (and the tests), that the traversal logic is still the same. And we still have the non-optimized search methods to understand the algorithm.

src/rust/src/http_reader.rs
pka: I prefer to go with the values you fund and decide later which values should be tunable.

Is `assumed_branching_factor` the same as `PackedRTree::DEFAULT_NODE_SIZE`?

pka (commented):
This really looks like a big improvoment, thanks a lot for digging into this!

Will do a few tests with small and bigger datasets.

src/rust/src/http_client.rs
michaelkirk: While tracking down the `stale tail` bug, I re-wrote this line in a way that made more sense to me. 

It's mathematically equivalent, but unnecessary. I'll revert it momentarily.

src/rust/src/http_reader.rs
michaelkirk: Ah yes, I'll reference that constant instead. 

The idea is that it would eventually be configurable, so if I know I'm accessing a FGB with branching factor of 32, I can get enough (and no extra) data.

src/rust/src/packed_r_tree.rs
michaelkirk: This is indeed the crux of the change and the barer of the bulk of complexity.

I'll add a high level comment to explain, as I should have done initially...

But basically:

Previously, each node traversed in the index tree was added separately to the search queue.

Now, rather than adding a new node, we'll merge it with an existing node provided:
1. there's already a node for that level of the index
2. that node is "reasonably" close (e.g if I somehow need the fourth node in a level and the three-millionth node in the level I probably don't want to merge those into a single request because I'll need to fetch all the items in-between)

src/rust/src/http_reader.rs
michaelkirk: Right, basically - but this is an intentional over-estimate since we don't yet have the real `item_count`.

If the actual item_count happened to be a power of 16, then I believe they should match exactly.

src/rust/src/http_client.rs
michaelkirk: Just to be clear, the bug fixed was the line before that: 

```
            let range_begin = max(begin, tail);
```

Is invalid if we've reset head (e.g. https://github.com/bjornharrtell/flatgeobuf/pull/93/files/003a76442ec9366491eaa1ed6fc1097890ca9c42#diff-75c21172adb6c41e0eb130a61d0bdd00983afafab292cfbc811844d964f21abeR69), then `tail` is no longer valid.

src/rust/src/packed_r_tree.rs
michaelkirk: Hopefully this is a little clearer with https://github.com/bjornharrtell/flatgeobuf/pull/93/commits/eedd2df84f4093d797d6e87d3129fe14a732d87d

I combined the second and third cases and tried to clarify that the only difference is what they log.

michaelkirk (commented):
I believe I've addressed your feedback and rerun my tests. PTAL @pka

pka:
Thanks for the explanations! My tests look good, too. Could you remove the "DO NOT MERGE" commit, so that I can rebase and merge?

pka:
Comparison reading header + index for same bbox of OSM building polygons.

osm-buildings-ch.fgb:
- Before: 6 requests, 4'195'336 bytes
- After: 5 requests, 640'144 bytes

osm-buildings-zurich.fgb:
- Before: 3 requests, 842'616 bytes
- After: 3 requests, 696'296 bytes

This confirms the big improvement especially for FGB files with many features and therefore a large index.

michaelkirk:
> Could you remove the "DO NOT MERGE" commit, so that I can rebase and merge?

Ok - Done! This should be good to merge.

bjornharrtell:
@michaelkirk excellent work! I hope to some day be inspired by it for other implementations.

PR #98, opened 2021-01-20
Rework TS implementation towards flatbuffers 2.x

bjornharrtell:


bjornharrtell:
Already done in master

PR #103, opened 2021-01-23
Fix multipolygon conversion 

xfischer:
Multipolygon with more than one interior ring (more than 2 ends) was causing error when reading


bjornharrtell (approved):

bjornharrtell:
@xfischer thanks 👍

PR #105, opened 2021-01-26
[ẂIP] Update to Rust flatbuffers 0.8.2

pka:
Tests fail when reading header from `countries.fgb` with error `Unaligned { position: 84, unaligned_type: "f64", error_trace: ErrorTrace([TableField { field_name: "envelope", position: 48 }]) }`


pka:
@bjornharrtell Do you have an idea why these `Unaligned` errors occur?

pka:
After looking deeper into flatbuffers, I'm pretty sure that this a a bug of the new Verifier with our FlatBuffers schema. They have a lot of [tests](https://github.com/google/flatbuffers/blob/master/tests/rust_usage_test/tests/integration_test.rs) with the demo schema, but seems like something in the FGB schema causes these "unaligned" error messages. I've tested with flatbuffers 0.8.2 and 0.8.2 und flatc from master.

bjornharrtell:
Nice work @pka, I had no other ideas either. But strange, the FlatGeobuf is nothing out of the ordinary.. might it be something with use of size prefix?

bjornharrtell:
@pka want to bring attention to this from flatbuffers people but unfortunately we probably/might need to have something more easily reproducible and possibly also proof that corresponding C++ implementation doesn't fail.

pka:
Still trying to reproduce the binary represention where the verifier errors occur. These files were (at least partially) produced with GDAL, but I currently try to reproduce it with Rust outputs of older flatc versions. The interesting part is that the FGB files are decodable, despite the verifier errors.

aardappel:
@CasperN
If this is indeed serious, open an issue on the FlatBuffers repo.. and may need to be a blocker for https://github.com/google/flatbuffers/issues/6353

CasperN:
Yep this seems serious and _something_ is wrong. An example that reproduces the error will be super helpful.

> The interesting part is that the FGB files are decodable, despite the verifier errors.

I will point out that `Unaligned { position: 444, unaligned_type: "f64", ...` does seem like its correctly identifying unaligned data `444 % 8 == 4`. This may have uncovered a latent bug in how data was produced, as Flatbuffers promises aligned data (at least w.r.t buffer[0]). That said, on CPUs that are cool with unaligned access, unless the compiler did something really fancy assuming pointers were aligned, errors don't manifest.

We can add a `check_alignment` option to the `VerfifierOptions` to opt out of alignment verification (this already exists in C++).

bjornharrtell:
@CasperN hmm I'm pretty sure the same binary passes the C++ verifier.

pka:
@CasperN Could the 4 byte size prefix cause the aligment error? Flatbuffer data may be correctly aligned relative to its beginning, but misaligned when reading with a 4 byte size prefix.
```rust
let mut f = File::open("../../test/data/triangle-header.fgb").unwrap();
let mut buf = Vec::new();
f.read_to_end(&mut buf).unwrap();
println!("{:?}", &buf);
// [116, 0, 0, 0, 32, 0, 0, 0, 0, 0, 0, 0, 0, 0, 22, 0, 28, 0, 8, 0, 12, 0, 7, 0, 0, 0, 0, 0, 0, 0, 0, 0, 16, 0, 20, 0, 22, 0, 0, 0, 0, 0, 0, 17, 60, 0, 0, 0, 20, 0, 0, 0, 12, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 4, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 34, 64, 0, 0, 0, 0, 0, 0, 34, 64, 8, 0, 0, 0, 116, 114, 105, 97, 110, 103, 108, 101, 0, 0, 0, 0]

let header = size_prefixed_root_as_header(&buf).unwrap();
println!("{:?}", &header);
assert_eq!(header.features_count(), 1);

let header = root_as_header(&buf[4..]).unwrap(); // FAILS with Unaligned { position: 68, unaligned_type: "f64", error_trace: ErrorTrace([TableField { field_name: "envelope", position: 44 }]) }
```

bjornharrtell:
@pka, a size prefixed flatbuffer is explictly requested from the builder to make it aligned with the size prefix included. Reading code must not assume it can move around the data after the size prefix.

bjornharrtell:
Ehm.. looking at my not so well maintained C++ code it doesn't look like I'm following how I remembered it in https://github.com/flatgeobuf/flatgeobuf/pull/105#issuecomment-793073000. So I'm a bit confused right now.. but I made the magicbytes header 8 bytes just because that is the largest datatype/alignment and I did have to resolve some alignment issues in implementing fx. the TypeScript deserializing.

CasperN:
> hmm I'm pretty sure the same binary passes the C++ verifier.

Is there an example in `flatgeobuf/test/data/`? I don't see `triangle-header.fgb`

bjornharrtell:
Hmm also in TypeScript I deliberately parse from 4 bytes offset in the buffer using the plain `getRoot` deserializer, see https://github.com/flatgeobuf/flatgeobuf/blob/master/src/ts/generic/featurecollection.ts#L187-L188. But this is also what the dedicated `getSizePrefixedRoot` does https://github.com/flatgeobuf/flatgeobuf/blob/9762163cd38e9add606782d1f0a2ecfad606d944/src/ts/feature_generated.ts#L81 so unless I'm misunderstanding here, it is the other way round - size prefixed buffers are assumed to be aligned at the 4 byte offset from the start of the buffer.

bjornharrtell:
@CasperN I believe @pka mentioned that this https://github.com/flatgeobuf/flatgeobuf/blob/master/test/data/countries.fgb file also doesn't pass the rust validator.

bjornharrtell:
@CasperN I see now that the smaller header only flatbuffer is at https://github.com/pka/flatgeobuf/blob/flatbuf-0.8/test/data/triangle-header.fgb.

pka:
Added [triangle-header.fgb](https://github.com/flatgeobuf/flatgeobuf/raw/621661ac9eef6a28ad5ab757e5f9781184a5665d/test/data/triangle-header.fgb).

I get alignement errors on all test files. My current explanation is, that all files are aligned including size prefix, but I decode them without size prefix.

bjornharrtell:
@pka it seems in GDAL where I use C++ code gen and the verifier I read the flatbuffer into memory starting at 0 *without* size prefix and then decode using the non-size-prefixed getroot:

Read:
https://github.com/OSGeo/gdal/blob/72891ee01fa59150050310253575430198f270b4/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp#L1413

Decode:
https://github.com/OSGeo/gdal/blob/72891ee01fa59150050310253575430198f270b4/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp#L1425

bjornharrtell:
Hmm I recalled a discussion about prefix and alignment and @aardappel said this:

image

And I had forgot but I even contributed the size prefix support for JS code-gen (https://github.com/google/flatbuffers/pull/5326) so I could avoid rolling my own size prefix which *would* have alignment issues.

But with this here issue I'm back to 100% confused again about this. :(

bjornharrtell:
And some more review of the C++ code I made for GDAL reveals that I had not understood this fully because there are two verifier functions available one that is size prefix aware one that is not and I use the latter at https://github.com/OSGeo/gdal/blob/72891ee01fa59150050310253575430198f270b4/gdal/ogr/ogrsf_frmts/flatgeobuf/ogrflatgeobuflayer.cpp#L1419. I'm at a loss why this has not become an issue much earlier.

aardappel:
@bjornharrtell The correct way that languages deal with a size pre-fix is to take it into account for buffer alignment just once, i.e. you compute which of { size_prefix, root_offset, file_identifier } you intend to write, then you align such that the final buffer after all of these will end up with `min_align` alignment (often 8). As always, in case of doubt, blindly copy the logic of the C++ implementation :)

pka:
According to latest test results, the C++ generated FGB files are aligned _including_ size prefix (so not aligned to 8 bytes from the start of the file) and the Rust generated FGB files are aligned _without_ size prefix. In Rust the difference is mady by using [finish](https://docs.rs/flatbuffers/0.8.3/flatbuffers/struct.FlatBufferBuilder.html#method.finish) or [finish_size_prefixed](https://docs.rs/flatbuffers/0.8.3/flatbuffers/struct.FlatBufferBuilder.html#method.finish_size_prefixed).

bjornharrtell:
@pka ok that seems logical at least, you should be using `finish_size_prefixed`.

I have added some unit tests for C++ verification and the results is however confusing me even more:

https://github.com/flatgeobuf/flatgeobuf/blob/79ba5b8c73a349509e9d579a3db25e770814cbaa/src/cpp/test/header.h#L19-L32

I do not understand how v1 can pass.. @aardappel can you explain this?

CasperN:
so, I tried reading with header.fbs in Rust and without size prefix for
* https://github.com/pka/flatgeobuf/blob/flatbuf-0.8/test/data/triangle-header.fgb
* https://github.com/flatgeobuf/flatgeobuf/blob/master/test/data/countries.fgb

Triangle works, its size prefixed

Countries does not work. It fails verification and then goes OOB if I override it with unsafe. I also tried ignoring alignment errors, the verifier does correctly predict OOB if not alignment errors.

```rust
use std::io::Read;                                                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                            
use flatbuffers;                                                                                                                                                                                                                                                                                                                                                            
mod header_generated;                                                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                            
fn main() {                                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                            
    let mut f = std::fs::File::open("countries.fgb").unwrap();                                                                                                                                                                                                                                                                                                              
    // let mut f = std::fs::File::open("triangle-header.fgb").unwrap();                                                                                                                                                                                                                                                                                                     
    let mut b = Vec::new();                                                                                                                                                                                                                                                                                                                                                 
    f.read_to_end(&mut b).unwrap();                                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                            
    let opts = flatbuffers::VerifierOptions {                                                                                                                                                                                                                                                                                                                               
        ignore_misalignment: true,                                                                                                                                                                                                                                                                                                                                          
        ..Default::default()                                                                                                                                                                                                                                                                                                                                                
    };                                                                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                             
    dbg!(flatbuffers::root_with_opts::<header_generated::Header>(&opts, &b));                                                                                                                                                                                                                                                                                               
    dbg!(flatbuffers::size_prefixed_root_with_opts::<header_generated::Header>(&opts, &b));                                                                                                                                                                                                                                                                                 
    unsafe {                                                                                                                                                                                                                                                                                                                                                                
        dbg!(flatbuffers::root_unchecked::<header_generated::Header>(&b));                                                                                                                                                                                                                                                                                                  
        dbg!(flatbuffers::size_prefixed_root_unchecked::<header_generated::Header>(&b));                                                                                                                                                                                                                                                                                    
    }                                                                                                                                                                                                                                                                                                                                                                       
}    
```

pka:
@CasperN `countries.fgb` starts with an 8 byte file signature. See file layout on https://flatgeobuf.org/#specification

CasperN:
After seeking 8 bytes, I was able to read the file using size_prefixed_root without having to opt out of alignment errors.

I assume this data looks correct, so I guess I still need more context to reproduce the errors above.

```
[src/main.rs:25] flatbuffers::size_prefixed_root::<header_generated::Header>(&b) = Ok(
    Header {
        name: Some(
            "countries",
        ),
        envelope: Some(
            [
                -180.0,
                -85.609038,
                180.0,
                83.64513,
            ],
        ),
        geometry_type: MultiPolygon,
        hasZ: false,
        hasM: false,
        hasT: false,
        hasTM: false,
        columns: Some(
            [
                Column {
                    name: "id",
                    type_: String,
                    title: None,
                    description: None,
                    width: -1,
                    precision: -1,
                    scale: -1,
                    nullable: true,
                    unique: false,
                    primary_key: false,
                    metadata: None,
                },
                Column {
                    name: "name",
                    type_: String,
                    title: None,
                    description: None,
                    width: -1,
                    precision: -1,
                    scale: -1,
                    nullable: true,
                    unique: false,
                    primary_key: false,
                    metadata: None,
                },
            ],
        ),
        features_count: 179,
        index_node_size: 16,
        crs: Some(
            Crs {
                org: Some(
                    "EPSG",
                ),
                code: 4326,
                name: Some(
                    "WGS 84",
                ),
                description: None,
                wkt: Some(
                    "GEOGCRS[\"WGS 84\",DATUM[\"World Geodetic System 1984\",ELLIPSOID[\"WGS 84\",6378137,298.257223563,LENGTHUNIT[\"metre\",1]]],PRIMEM[\"Greenwich\",0,ANGLEUNIT[\"degree\",0.0174532925199433]],CS[ellipsoidal,2],AXIS[\"latitude\",north,ORDER[1],ANGLEUNIT[\"degree\",0.0174532925199433]],AXIS[\"longitude\",east,ORDER[2],ANGLEUNIT[\"degree\",0.0174532925199433]],ID[\"EPSG\",4326]]",
                ),
                code_string: None,
            },
        ),
        title: None,
        description: None,
        metadata: None,
    },
)
```

pka:
@CasperN The Rust verfier errors were caused by decoding the C++ output without size prefix. After changing that in the reader, the Rust side is ok!

I think, the only open question are some unexpected C++ test results: https://github.com/flatgeobuf/flatgeobuf/pull/105#issuecomment-793131084

src/rust/src/file_reader.rs
```diff
@@ -29,7 +29,10 @@ impl<'a, R: Read + Seek> FgbReader<'a, R> {
     pub fn open(reader: &'a mut R) -> Result<Self> {
         let mut magic_buf: [u8; 8] = [0; 8];
         reader.read_exact(&mut magic_buf)?;
-        if magic_buf != MAGIC_BYTES {
+        if magic_buf[0..2] != MAGIC_BYTES[0..2]
```
michaelkirk, line 32:
```suggestion
        if magic_buf[0..=2] != MAGIC_BYTES[0..=2]
```

This part seems off (and similarly below).

src/rust/src/http_reader.rs
```diff
@@ -58,15 +60,19 @@ impl HttpFgbReader {
         if bytes != MAGIC_BYTES {
```
michaelkirk, line 60: Do we need a more permissive version check here, like you did with the file reader?

michaelkirk (commented):
I don't have time for a full review today, but I'd like to do so tomorrow.

src/rust/src/http_reader.rs
```diff
@@ -85,7 +91,7 @@ impl HttpFgbReader {
         self.fbs.header()
     }
     fn header_len(&self) -> usize {
-        12 + self.fbs.header_buf.len()
+        8 + self.fbs.header_buf.len()
```
michaelkirk, line 94: I think there is logic elsewhere that assumes this length includes the length i32 is included, so I think you'll also need to update some users, like `select_bbox`.

src/rust/src/http_reader.rs
pka: Yes, we should have the same. Forgot it.

src/rust/src/http_reader.rs
pka: Not sure I understand. `header_len` *does* still include the length i32, but the length i32 is now included in `fbs.header_buf`

michaelkirk: Oh, my mistake! Thanks for clarifying.

src/rust/src/http_reader.rs
```diff
@@ -145,14 +151,18 @@ impl HttpFgbReader {
...
-            .get_range(self.pos, feature_size, min_req_size)
-            .await?;
+        let feature_size = LittleEndian::read_u32(&bytes) as usize;
+        bytes.put(
+            self.client
+                .get_range(self.pos, feature_size, min_req_size)
+                .await?,
+        );
         self.fbs.feature_buf = bytes.to_vec(); // Not zero-copy
+                                               // verify flatbuffer
```
michaelkirk, line 163:
```suggestion
        // verify flatbuffer
```

nit: I think this comment applies to the subsequent line, so this is a weird indent 🤓

src/rust/src/properties_reader.rs
```diff
@@ -9,19 +9,21 @@ use std::str;
 
 /// Access to current feature
 pub struct FgbFeature {
-    pub(crate) header_buf: Vec<u8>,
+    pub(crate) header_buf: Vec<u8>, // Header would require lifetime
```
michaelkirk, line 12: Can you clarify this comment?

michaelkirk (approved):
Other than the comments I made last week re: version checking, everything LGTM.

I did a sanity check running the updated HTTPReader against an old FGB file, and things seem reasonable.

Nice work!

src/rust/src/http_reader.rs
pka: I blaim `cargo fmt` for that....

src/rust/src/properties_reader.rs
pka: Would `Type "Header" would require lifetime for struct` be understandable?

michaelkirk: Sorry, I'm still not sure what the comment is trying to tell me. 


Is it that you are trying to document that `header()` borrows this `header_buf` field?

```
let some_fgb_feature: FgbFeature = get_fgb_feature();

// __this call borrows `header_buf` with lifetime `a __
let header: pub struct Header<'a> = some_fgb_feature.header();
```

michaelkirk: Ah, I think I get it.

You're saying that you chose for FgbFeature to have a `header_buf: Vec<u8>` field rather than a `header: Header` field because having a `Header` field would require FgbFeature to cary a generic lifetime, like FgbFeature<'a>. 

Is that it?

src/rust/src/lib.rs
```diff
@@ -107,6 +107,12 @@ pub use fallible_streaming_iterator::FallibleStreamingIterator;
...
 pub const VERSION: u8 = 3;
-pub const MAGIC_BYTES: [u8; 8] = [b'f', b'g', b'b', VERSION, b'f', b'g', b'b', 0];
+const MAGIC_BYTES: [u8; 8] = [b'f', b'g', b'b', VERSION, b'f', b'g', b'b', 0];
 
-pub const HEADER_MAX_BUFFER_SIZE: usize = 1048576 * 10;
+const HEADER_MAX_BUFFER_SIZE: usize = 1048576 * 10;
+
+fn check_magic_bytes(magic_bytes: &[u8]) -> bool {
+    magic_bytes[0..3] == MAGIC_BYTES[0..3]
+        && magic_bytes[4..7] == MAGIC_BYTES[4..7]
```
michaelkirk, line 116:
```suggestion
        && magic_bytes[4..8] == MAGIC_BYTES[4..8]
```

Do we need to veritfy the final `0` as well?

pka: I guess you're right. Despite being added just for padding, the trailing zero is part of the magic bytes.

PR #107, opened 2021-02-18
For the Rust crate, disable default features of reqwest. It brings in

dabreegster:
system bindings to SSL, which are undesirable for some users. Since
features are additive, this frees up callers to make the choice instead.

This is for a-b-street/abstreet/issues/519

### DIFF
--- a/src/rust/Cargo.toml
+++ b/src/rust/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "flatgeobuf"
-version = "0.4.1"
+version = "0.5.0"
 authors = ["Pirmin Kalberer <pka@sourcepole.ch>"]
 edition = "2018"
 description = "FlatGeobuf for Rust."
--- a/src/rust/Cargo.toml
+++ b/src/rust/Cargo.toml
@@ -16,7 +16,7 @@ flatbuffers = "0.6"
 byteorder = "1.4.2"
 geozero = "0.6.0"
 async-trait = "0.1.42"
-reqwest = "0.11.0"
+reqwest = { version = "0.11.0", default-features = false }
 bytes = "1.0.1"
 log = "0.4.13"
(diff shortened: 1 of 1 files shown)


dabreegster:
@pka, @michaelkirk 

I think this is a breaking API change, so should I bump to... 0.4.2? Or 0.5? I'll read about semver rules.

dabreegster:
https://doc.rust-lang.org/cargo/reference/semver.html#cargo-change-dep-feature
I think this is considered a major change, because somebody before might've been using flatgeobuf without specifying `reqwest` features any other way. After this PR, if they happened to be fetching HTTPS, they'd break. So I changed the version to 0.5.

pka (approved):
:+1:

pka:
It's probably a good moment to add a feature flag for the http-client, since its still adding quite a few dependencies.

bjornharrtell:
Hey @dabreegster nice work on A/B Street and happy to see you have found flatgeobuf of use.

dabreegster:
> It's probably a good moment to add a feature flag for the http-client, since its still adding quite a few dependencies.

Good idea! I can get to this... eventually. :(

> Hey @dabreegster nice work on A/B Street and happy to see you have found flatgeobuf of use.

Thanks! @michaelkirk is using it to read just a few census tracts out of a massive file, so we can generate more realistic traffic patterns. It's a huge help -- without it, we'd have to chop up all the census data into smaller files, make decisions about how to split at state/county boundaries, deal with a map on state borders pulling from multiple files, etc. Huge headache avoided.

michaelkirk:
>> It's probably a good moment to add a feature flag for the http-client, since its still adding quite a few dependencies.
Good idea! I can get to this... eventually. :(
>
> Good idea! I can get to this... eventually. :(

Already done by @pka in https://github.com/flatgeobuf/flatgeobuf/pull/108!

PR #108, opened 2021-02-20
Make http an optional feature

pka:
Dependency tree without http feature:

```
$ cargo tree --no-default-features --edges no-dev
flatgeobuf v0.5.0 (/home/pi/code/gis/flatgeobuf/src/rust)
├── byteorder v1.4.2
├── fallible-streaming-iterator v0.1.9
├── flatbuffers v0.6.1
│   └── smallvec v1.6.1
├── geozero v0.6.0
│   ├── async-trait v0.1.42 (proc-macro)
│   │   ├── proc-macro2 v1.0.24
│   │   │   └── unicode-xid v0.2.1
│   │   ├── quote v1.0.9
│   │   │   └── proc-macro2 v1.0.24 (*)
│   │   └── syn v1.0.60
│   │       ├── proc-macro2 v1.0.24 (*)
│   │       ├── quote v1.0.9 (*)
│   │       └── unicode-xid v0.2.1
│   ├── seek_bufread v1.2.2
│   └── thiserror v1.0.23
│       └── thiserror-impl v1.0.23 (proc-macro)
│           ├── proc-macro2 v1.0.24 (*)
│           ├── quote v1.0.9 (*)
│           └── syn v1.0.60 (*)
└── log v0.4.14
    └── cfg-if v1.0.0
```

### DIFF
--- a/src/rust/tests/http_read.rs
+++ b/src/rust/tests/http_read.rs
@@ -1,84 +1,88 @@
-use flatgeobuf::*;
-use geozero::error::Result;
-use tokio::runtime::Runtime;
+#[cfg(feature = "http")]
+mod http {
 
-async fn http_read_async() -> Result<()> {
-    let url = "https://github.com/flatgeobuf/flatgeobuf/raw/master/test/data/countries.fgb";
-    let mut fgb = HttpFgbReader::open(url).await?;
-    assert_eq!(fgb.header().geometry_type(), GeometryType::MultiPolygon);
-    assert_eq!(fgb.header().features_count(), 179);
-    fgb.select_all().await?;
-    let feature = fgb.next().await?.unwrap();
-    let props = feature.properties()?;
-    assert_eq!(props["name"], "Antarctica".to_string());
-    Ok(())
-}
+    use flatgeobuf::*;
+    use geozero::error::Result;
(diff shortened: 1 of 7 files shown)


src/rust/tests/http_read.rs
```diff
@@ -1,84 +1,88 @@
-use flatgeobuf::*;
-use geozero::error::Result;
-use tokio::runtime::Runtime;
+#[cfg(feature = "http")]
+mod http {
```
michaelkirk, line 2: Just a style nit, but you could hoist this condition to the outer `mod http_read` line.

michaelkirk (approved):
This LGTM. 

I tested that the http reader works for both the native and wasm builds of http://github.com/a-b-street/abstreet

src/rust/tests/http_read.rs
pka: Since this under `tests`, there isn't an explicit `mod http_read` line. If there is a more elegant way to disable tests with feature flags, I'm interested!

michaelkirk: oh my mistake! I don't know of another way.

bjornharrtell (approved):

PR #110, opened 2021-03-04
Java serialization: memory and speed improvements

groldan:
Hi all,

I've been testing the GeoServer WFS output format with the docker composition from https://github.com/camptocamp/geoserver-cloud.

Had the WFS containers capped at 2 CPUs with 1GB of RAM with 80% available to the JVM, and they were exiting with OOM.

So increased the container memory to 2GB with 50% for the JVM, and noticed there's still quite a spike in memory consumption.

Put hands-on and got these improvements, that reduce the memory footprint of Java serialization and increases the throughput.

The commit messages should be informative enough of what's going on, so I won't repeat here.

Bottom line, here are some results:

|   | Max Throughput (reqs/s) | Memory | Memory Delta (*) |
| ------------- | ------------- | ----------- | ----------- |
| GML2 Baseline | ~530  | ~700 | ~200 |
| FBS Baseline  | ~130  | +1.2GiB | ~700 |
| FBS Improvements  | > 1000  | ~600 | ~100 |

> (*) Taking into account that the initial memory usage is around 500MiB (container memory, doesn't mean it's all JVM's Heap)

That's close to an order of magnitude in performance, at about 1/7 of memory footprint.

Here are some screencasts of what it looks like. 
The top half of the screen shows the output of `docker stats`, watch out the `MEM USAGE / LIMIT` columns for the `gs_cloud_wfs_X` containers as I run these simple tests:

`test_gml2.sh`:
```bash
#!/bin/bash
fbs="http://localhost:9090/wfs?request=GetFeature&typeName=topp:states&outputFormat=gml2"
for i in 2 4 8 16 32; do echo "Concurrency: $i" && ab -t 3 -c $i "$fbs" 2>/dev/null|grep -e "Requests per second";done
```

`test_fbs.sh`:
```bash
#!/bin/bash
fbs="http://localhost:9090/wfs?request=GetFeature&typeName=topp:states&outputFormat=application/flatgeobuf"
for i in 2 4 8 16 32; do echo "Concurrency: $i" && ab -t 3 -c $i "$fbs" 2>/dev/null|grep -e "Requests per second";done
```

## GML Baseline

https://user-images.githubusercontent.com/207423/109921883-4b4c5180-7c9b-11eb-8b51-c77ff8998a6a.mp4
> GML2 baseline (fastest GML encoder in GeoServer)

## FlatGeoBuff Baseline

https://user-images.githubusercontent.com/207423/109923853-29080300-7c9e-11eb-9505-22b89342c5cb.mp4
> FlatGeoBuff baseline (master at f08c2138bf5bb89e2316fc70c7f07f0a5f64ff96)

## FlatGeoBuff Improvements

https://user-images.githubusercontent.com/207423/109922302-e2190e00-7c9b-11eb-891c-e2c6b5f0bcb5.mp4
> FlatGeoBuff improvements (this PR at 6f0732cf1ee5c816f8a7dcc34bebc2b9db49b2b8)

---
Although I didn't write new unit tests, all existing ones pass, and I've checked the output of GeoServer is exactly the same as the baseline's.

### DIFF
--- a/src/java/pom.xml
+++ b/src/java/pom.xml
@@ -92,6 +92,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-javadoc-plugin</artifactId>
+        <version>3.2.0</version>
         <executions>
           <execution>
             <id>attach-javadocs</id>
@@ -100,6 +101,15 @@
             </goals>
           </execution>
         </executions>
+        <configuration>
+          <tags>
+            <tag>
+              <name>implNote</name>
+              <placement>a</placement>
+              <head>Implementation Note:</head>
+            </tag>
+          </tags>
+        </configuration>
       </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
--- a/src/java/src/main/java/org/wololo/flatgeobuf/geotools/FeatureCollectionConversions.java
+++ b/src/java/src/main/java/org/wololo/flatgeobuf/geotools/FeatureCollectionConversions.java
@@ -5,6 +5,7 @@ import java.io.OutputStream;
 import java.nio.ByteBuffer;
 
 import com.google.flatbuffers.ByteBufferUtil;
+import com.google.flatbuffers.FlatBufferBuilder;
 import static com.google.flatbuffers.Constants.SIZE_PREFIX_LENGTH;
 
 import org.wololo.flatgeobuf.generated.Feature;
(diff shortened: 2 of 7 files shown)


bjornharrtell:
@groldan very nice work! All is clear except I have problems quickly understanding how/why https://github.com/flatgeobuf/flatgeobuf/pull/110/commits/6f0732cf1ee5c816f8a7dcc34bebc2b9db49b2b8 works. But I will merge this and make a relase anyway because this is clearly very good improvements to the Java ref impl. Very much appreciated, thanks you!

bjornharrtell:
@groldan released as 3.10.0 and is up at maven central - https://repo1.maven.org/maven2/org/wololo/flatgeobuf/3.10.0/.

bjornharrtell:
Also made a https://repo1.maven.org/maven2/org/wololo/flatgeobuf/3.10.1/ with updated geotools dep and jts made transitive.

groldan:
Excellent! Thank you

groldan:
> I have problems quickly understanding how/why [6f0732c](https://github.com/flatgeobuf/flatgeobuf/commit/6f0732cf1ee5c816f8a7dcc34bebc2b9db49b2b8) works.

BTW, that's doing the same than `Geometry.createXyVector(FlatBufferBuilder , double[])`:

```java
  public static int createXyVector(FlatBufferBuilder builder, double[] data) {
      builder.startVector(8, data.length, 8);
      for (int i = data.length - 1; i >= 0; i--)
          builder.addDouble(data[i]);
      return builder.endVector();
  }
```

Except that we don't create an in memory `double[]` as an intermediate step.
Instead, it traverses the JTS geometry ordinates in reverse order just like that method does, calling `builder.addDouble()` for each ordinate. The `CoordinateSequenceFilter` does it for each geometry's coordinate sequence, and the `applyInReverseOrder()` method does it for each geometry member, so that at no point we need to create extra objects in memory, either `Coordinate` or `double[]`.

PR #113, opened 2021-04-05
java: write layer name

cportele:
It seems as if the name is set in the header, the name is not written and `ogrinfo` always shows "unknown".


bjornharrtell:
Nice catch.

PR #114, opened 2021-04-25
fix path to fgb in example

michaelkirk:
https://flatgeobuf.org/examples/leaflet/index_filtered.html currently 404's when fetching the fgb. 

This should fix it. I tested that /examples/leaflet/index.html and /examples/leaflet/index_filtered.html both work locally after this change.


bjornharrtell (approved):
Thanks!

PR #116, opened 2021-05-03
js: fix reading from files larger than 4GB

michaelkirk:
`offset` is a 64bit number, but we were only reading the lower 32, so things break when reading features from a file larger than 4GB.

This PR only adds support for up to 52 bit integers, "because javascript", but I think it's worth doing. 

In my case I'm filtering a ~15GB dataset down to a very small subset, so the total data transferred is a handful of MB.


bjornharrtell (approved):
Makes sense!

Issue #50, opened 2020-04-01
Optimize I/O for JavaScript implementation


bjornharrtell:
Prior art is the general prefetch strategy in GDAL, see https://gdal.org/user/virtual_file_systems.html#vsicurl-http-https-ftp-files-random-access.

kylebarron:
Is your idea to retrieve the beginning metadata and then potentially use range requests to read portions of the rest of the file? That's a strategy used for Cloud Optimized Geotiffs.

bjornharrtell:
The current implementation already does that on filtered access but a bit too naively, see https://github.com/bjornharrtell/flatgeobuf/blob/0ce5c837491a41a0f6771ab150bbd7eb0e9a8c26/src/ts/generic/featurecollection.ts#L87-L91.

The problem with the naive approach is that it can result in many small requests. What this issue is about is about using a generic prefetch strategy to reduce the amount of requests, which is what GDAL does.

pka:
The performance win for buffered http access is huge! In the Rust driver I've added a buffer to the http client and an additional `get` method with a `min_req_size` parameter:
https://github.com/bjornharrtell/flatgeobuf/blob/master/src/rust/src/http_reader.rs#L59
It's currently set to 512B for header access, the full index size for index access and to 1MB for data access.

bjornharrtell:
@pka that `BufferedHttpClient` seems very clever. :)

michaelkirk:
Some related work for the rust client in #93

michaelkirk:
After fixing the bbox'd example in #114, I was surprised to see that FGB made 50 requests to render the 17 features.

<img width="2032" alt="Screen Shot 2021-04-24 at 7 04 25 PM" src="https://user-images.githubusercontent.com/217057/115977782-f0630800-a52f-11eb-833b-942d0067e6bb.png">

Is that surprising, and covered by the work you anticipated doing here, or do you suspect that is another bug and I should open a separate issue?

bjornharrtell:
@michaelkirk current implementation is very naive about I/O so it is not unexpected and it is what this issue is about.

michaelkirk:
Ok, that makes sense, just wanted to make sure.

Would you be interested in a typescript implementation similar to what I did for rust in #93? This is something I'd be interested in working on.

bjornharrtell:
Absolutely, much appreciated!

michaelkirk:
[...]

PR #117, opened 2021-05-04
js: more efficient Bbox search

michaelkirk:
FIXES: https://github.com/flatgeobuf/flatgeobuf/issues/50

- Introduce a buffered reader, like @pka introduced for Rust

- Fetch entire header + tree prefix with first 12kb request to obviate
  the first few index requests  (like in #93)
- Merge subsequent nearby index requests into a single request (like in #93)
- Fetch extra buffered feature data to obviate some requests (like in #93)

**before:**
Screen Shot 2021-05-04 at 2 43 23 PM

**after:**
Screen Shot 2021-05-04 at 2 42 33 PM

Note: we're fetching more data after this change, but with substantially less requests.


src/ts/Logger.ts
```diff
@@ -0,0 +1,48 @@
+export enum LogLevel {
+    Debug, Info, Warn, Error
+}
+
+export default class Logger {
+    static logLevel: LogLevel = LogLevel.Info;
```
michaelkirk, line 6: You'll notice I left in bit of logging. I could have deleted it all before opening the PR, but if we want to come back to modify or debug this code later, it's helpful if the logging is already in place and can be easily toggled.

There's a reason logging typically works like this basically everywhere...

This line is where the logLevel is currently set - so none of the existing log statements will output anything without modifying this.

I'm not familiar with the JS world. Is there a more conventional way to do conditional logging like this in JS?

src/ts/packedrtree.ts
```diff
@@ -84,13 +151,40 @@ export async function* streamSearch(
...
+            if (tail !== undefined && tail.level() == next.level() - 1) {
+                Logger.debug(`pushing new node at offset: ${offset} rather than merging with distant ${tail}`);
+            } else {
+                Logger.debug(`pushing new level for ${newNodeRange} onto queue with tail: ${tail}`);
+            }
+
+            queue.push(newNodeRange);
         }
-        // order queue to traverse sequential
-        queue.sort((a, b) => b[0] - a[0])
```
michaelkirk, line 93: Since we now `shift` the next node off, rather than `pop` it, our `queue` is ordered by construction, and this `sort` call became unnecessary.

src/ts/HttpReader.ts
```diff
@@ -0,0 +1,220 @@
...
+
+import { Rect, calcTreeSize, DEFAULT_NODE_SIZE, NODE_ITEM_LEN, streamSearch} from './packedrtree';
+import { magicbytes, SIZE_PREFIX_LEN } from './constants';
+import Logger from './Logger';
+import HeaderMeta from './HeaderMeta'; 
+import { Feature } from './feature_generated';
+
+export class HttpReader {
+    private headerClient: BufferedHttpRangeClient;
+    private _featureClient?: BufferedHttpRangeClient;
```
michaelkirk, line 11: One place we diverged from the rust client...

In rust we parse the entire index, then start parsing features.
whereas in JS we were yielding a feature as soon as we got one, while still working through the index.

The repercussion there is that we'll be jumping back and forth between reading the index and reading the features, so we need separate buffers.

src/ts/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const magicbytes: Uint8Array = new Uint8Array([0x66, 0x67, 0x62, 0x03, 0x66, 0x67, 0x62, 0x00])
```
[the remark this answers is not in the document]
bjornharrtell: 👍

src/ts/HeaderMeta.ts
```diff
@@ -14,4 +17,25 @@ export default class HeaderMeta {
         public metadata: string | null
         ) {
     }
+
+    static fromByteBuffer(bb: flatbuffers.ByteBuffer) : HeaderMeta {
```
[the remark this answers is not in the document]
bjornharrtell: 👍

src/ts/Logger.ts
bjornharrtell: I've used https://github.com/pinojs/pino before but not sure it's warranted to bring that in here.

src/ts/packedrtree.ts
bjornharrtell: Nice!

src/ts/HttpReader.ts
bjornharrtell: Not sure of the pros and cons here. I guess there could be cases where forward seek only could be desirable but I suspect that I/O pattern/requirement is disappearing fast. So I'm leaning towards accepting this.

bjornharrtell (commented):
Overall looks very good, I'll look into it some more tomorrow but I'll give a preliminary great work remark on this. 🙂

src/ts/Logger.ts
michaelkirk: Unless you have a clear preference, I'd prefer to just use this really minimal dumb thing for now rather than include an external dependency.

src/ts/HttpReader.ts
michaelkirk: FWIW, I mostly went this way just because its' what the JS code was already doing.

Some pros that come to mind with this approach though:
1. time to first feature is minimized
2. we don't have to hold all the search results (`Array<[byteOffsets, featureIdx]>`) in memory at once

cons:
1. two buffers means more memory... but it's a fixed amount rather than point (2.) above.
2. seek times as you mentioned, though I imagine once network latency is involved, that disk seek times are less significant.

src/ts/Logger.ts
bjornharrtell: We'll keep the dumb thing :)

src/ts/HttpReader.ts
bjornharrtell: Thanks for the explanation, I'm fine with keeping it as is.

bjornharrtell (approved):

Issue #118, opened 2021-05-04
revamp JS examples


michaelkirk:
- [ ] Include examples which run off the local build (i'm tired of having to always modify them to point to my local 😆)
- [x] make it easier to discover all the other examples - e.g. a meta-index which links to all of the examples.
  - full file
  - search
  - for both leaflet and open layers

Also, as the JS i/o gets better (#50), it might be neat to post slightly different examples:
- [x] search a *very large* file - I have such a file. Would you be interested in moving it to the fgb s3 bucket?
- [x] dynamic search (e.g. drag a bounding rect around)

michaelkirk:
I'm happy to do this work - just wanted to run it by y'all to see if you had any thoughts.

bjornharrtell:
Good ideas. I'll be happy to support/help out and if possible it would be great to put up an example of huge file bbox search.

tim-salabim:
If at all possible, it would be great to re-visit the leaflet example. @bjornharrtell pointed out to me once that there may be better implementations than the current example. We rely on this example quite heavily in the leafem R package. Even though the approach seems to work very nicely and is robust, I am always interested in improvements. Thanks again for all the great work here!!

michaelkirk:
@tim-salabim, for this particular issue, I'm only intending to work on showing off what the library can already do.

I think you might be talking about #23, which indeed sounds interesting!

tim-salabim:
@michaelkirk thanks for the heads-up. Yes, I am particularly interested in a performant leaflet binding for our very narrow use-case of data interchange between R and browser (for the vis). I'd like to emphazise though, that I am already very happy with the current implementation! So no worries really

bjornharrtell:
I have thrown togheter an example "openlayers-hnv". When zoomed in quite far it attempts to request bbox filtered from google cloud cloud storage from a 3.4 GB FlatGeobuf, unfortunately it's not going well. I get network requests like this:

![image](https://user-images.githubusercontent.com/141030/117180891-f4ffaa00-add4-11eb-89cc-28c238a90d26.png)

And it seem to continue indefinitely with 131 kB large bits.

Note that it uses a local build from master. I'll dig into it a bit when I have time, but wanted to give you heads up @michaelkirk.

bjornharrtell:
@michaelkirk nevermind! It was my mistake, the rect argument was all wrong.

bjornharrtell:
[...]

PR #120, opened 2021-05-06
local examples (including large and interactive filter demos)

michaelkirk:
Before I go further, I wanted to check in on the work I've done so far for #118 

Here's a video of the leaflet demos using the local build of the fgb libs:


examples/leaflet_local/filtered.html
```diff
@@ -0,0 +1,140 @@
...
+    </ul>
+    <ul class="secondary-navigation">
+        <li><a href="/examples/leaflet_local/">Basic Example</a></li>
+        <li class="active">Filter By Rect</li>
+        <li><a href="/examples/leaflet_local/large.html">Filtering a Large Dataset</a></li>
+    </ul>
+
+    <div id="map"></div>
+    <script>
+        document.addEventListener("DOMContentLoaded", async () => {
```
michaelkirk, line 31: For purposes of example's, I think it's just a little easier for people to see the script in-lined into the HTML, so everything they need to get started can be copy/pasted from one spot.

Is that alright?

examples/leaflet_local/filtered.html
```diff
@@ -0,0 +1,140 @@
...
+            updateResults();
+
+            // ...and update the results whenever the map moves
+            map.on("moveend", function(s){
+                rectangle.setBounds(getBoundForRect());
+                updateResults();
+            });
+        });
+    </script>
+    <p>
```
michaelkirk, line 118: I added a little copy to augment the examples.

examples/leaflet_local/index.html
```diff
@@ -0,0 +1,107 @@
...
+    <link rel="stylesheet" href="/examples/site.css" />
+    <script src="https://unpkg.com/leaflet@1.6.0/dist/leaflet.js"></script>
+    <script src="/dist/flatgeobuf-geojson.min.js"></script>
+    <script src="https://unpkg.com/json-formatter-js"></script>
+
+</head>
+<body>
+<style>
+</style>
+    <ul class="primary-navigation">
```
michaelkirk, line 13: I think it's helpful to have the examples link to each other... but for now that means copying and pasting around nav-menu snippets and slightly modifying them for each file to designate the active file.

Note these examples all read the local builds (src="/dist/flatgeobuf-geojson.min.js"), so if we duplicate these examples for a "hosted" version, we'd need to copy/paste/edit these around some more.

Alternatively, we could get rid of this duplication and do something slightly smarter by introducing some kind of static-site gen thing, but it'd mean introducing more tooling. Would you be comfortable with that?

I'd be happy either way, but I *do* think it's helpful to be able to navigate between the examples like this.

examples/leaflet_local/filtered.html
```diff
@@ -0,0 +1,140 @@
...
+                                weight: 4,
+                                fillOpacity: 0.7,
+                            });
+                            layer.bringToFront();
+                        },
+                        'mouseout': function(e) {
+                            const layer = e.target;
+                            layer.setStyle(defaultStyle);
+                        }
+                   }).bindPopup(`<h1>${feature.properties["NAME"]} ${feature.properties["LSAD"]}, ${feature.properties["STATE"]}</h1>`)
```
michaelkirk, line 98: I added a pop-up to show how you can read feature data from the FGB

examples/leaflet_local/large.html
```diff
@@ -0,0 +1,145 @@
...
+            async function updateResults() {
+                // remove the old results
+                previousResults.remove();
+                const nextResults = L.layerGroup().addTo(map);
+                previousResults = nextResults;
+
+                // Use flatgeobuf JavaScript API to iterate features as geojson.
+                // Because we specify a bounding box, flatgeobuf will only fetch the relevant subset of data,
+                // rather than the entire file.
+                let iter = flatgeobuf.deserialize('https://s3.amazonaws.com/mjk_asdf/abs/population_areas.fgb', fgBoundingBox(), handleHeaderMeta);
```
michaelkirk, line 74: This is my go-to 12.3GB FGB of United State census tracts. Is there an "official" FGB asset store that this could be copied to? Or perhaps another very large example FGB to show off?

michaelkirk: Also, it'd be more impressive if it was hosted behind a proper CDN.

examples/leaflet_local/filtered.html
bjornharrtell: Agreed I will change the existing ones to that.

examples/leaflet_local/index.html
bjornharrtell: Sure would be nice to have something similar to what openlayers have.

examples/leaflet_local/large.html
bjornharrtell: Don't have anything at the moment but I'll see what I can find.

examples/leaflet_local/index.html
michaelkirk: > something similar to what openlayers have.

Do you mean like this page: https://openlayers.org/en/latest/examples/ ?

bjornharrtell: Yeah.. it's quite nice. But we don't have to shoot for that level of UX as the first interation though.

examples/leaflet_local/large.html
bjornharrtell: And here it is :)

http://flatgeobuf.septima.dk/population_areas.fgb

bjornharrtell:
@michaelkirk the mp4 wouldn't show in GitHub but I found it and watched it now and it's very nice, will be a great example.

examples/leaflet_local/large.html
michaelkirk: done, thanks!

script/gen_examples.sh
```diff
@@ -0,0 +1,20 @@
+set -ex
+
+# Copies the dev examples into the `examples` root and replaces the local 
+# fgb libary with the published hosted one
```
michaelkirk, line 4: This script might be controversial... maybe you'll have a better suggestion.

The problem I'm trying to solve is that I want to be able to have a "local" version of the examples that I can dev on and which use the locally built FGB libraries.

...but I don't want to have to maintain twice as many example HTML files just for this small change — hence this script. It's admittedly pretty dumb. I am open to alternatives, or if you find it grossly offensive, I could just get rid of the local-dev versions and keep editing them each time I want to work locally.

Though a side benefit is, next time you release, you don't have to update all the examples to point to the latest release on unpkg.com, just this script. 😉

examples/openlayers/filtered.html
```diff
@@ -0,0 +1,154 @@
...
+                maxPoint.transform('EPSG:3857', 'EPSG:4326')
+
+                return {
+                    minX: minPoint.getCoordinates()[0],
+                    minY: minPoint.getCoordinates()[1],
+                    maxX: maxPoint.getCoordinates()[0],
+                    maxY: maxPoint.getCoordinates()[1],
+                };
+            }
+
```
michaelkirk, line 76: I spent way too much time trying to render the yellow bounding rect on openlayers like I did for leaflet. I gave up after a while. It'd be nice to have, but I think the example works OK without it.

michaelkirk:
Ok, this is ready for review. I did move some files around, but the only one that I know which is linked to, (/examples/leaflet/index.html) remains in the same spot. So hopefully no broken links.

bjornharrtell:
Would like to merge but I'm hesitant to have symbolic links in the repository, could it be avoided?

bjornharrtell:
Hmm on second thought. Considering I want to make a release soon and publish these examples publicly I'd like to find a better solution to have them in the repo with regard to be able to run them locally and publish externally without having to duplicate things.

michaelkirk:
> Would like to merge but 

> I'd like to find a better solution to have them in the repo with regard to be able to run them locally and publish externally without having to duplicate things.

How about I just delete the dev and symbolic link stuff for now and we could add that functionality if and when there's a solution you're comfortable with?

bjornharrtell:
> > Would like to merge but
> 
> > I'd like to find a better solution to have them in the repo with regard to be able to run them locally and publish externally without having to duplicate things.
> 
> How about I just delete the dev and symbolic link stuff for now and we could add that functionality if and when there's a solution you're comfortable with?

Yep let's do that

michaelkirk:
>> How about I just delete the dev and symbolic link stuff for now and we could add that functionality if and when there's a solution you're comfortable with?
>
> Yep let's do that

Ok done! 

Note that these examples will need to have their flatgeobuf version bumped when you release.

examples/openlayers/filtered.html
```diff
@@ -4,7 +4,7 @@
     <link rel="stylesheet" href="https://cdn.jsdelivr.net/gh/openlayers/openlayers.github.io@master/en/v6.5.0/css/ol.css" type="text/css">
     <script src="https://cdn.jsdelivr.net/gh/openlayers/openlayers.github.io@master/en/v6.5.0/build/ol.js"></script>
     <script src="https://unpkg.com/underscore@1.13.1/underscore-min.js"></script>
-    <script src="https://unpkg.com/flatgeobuf@3.7.1/"></script>
+    <script src="https://unpkg.com/flatgeobuf@3.7.1/dist/flatgeobuf-ol.min.js"></script>
```
michaelkirk, line 7: This was actually a bug in my website "build script", so I'm glad I had the opportunity for further scrutiny.

Issue #122, opened 2021-05-06
Further TypeScript I/O optimizations


bjornharrtell:
Some ideas for even more I/O optimization:

* [ ] Configurable minimum feature buffer size
* [ ] Buffer size doubling on subsequent requests (up to a default and configurable maximum)
* [ ] Optional cache/reuse of buffered data for header and index for separate requests against the same URL

Any opinions @michaelkirk?

michaelkirk:
>  Optional cache/reuse of buffered data for header and index for separate requests against the same URL

Currently the JS API is kind of a oneshot thing. I think to support this we might want to change the API to expose something like the HttpReader that the user can repeatedly `setBBox` on. What do you think? Or were you thinking some kind of global hidden cache?

> Buffer size doubling on subsequent requests (up to a default and configurable maximum)
> Configurable minimum feature buffer size

On a slightly related tangent, I have a question about the FGB format: Is it safe to assume that features are tightly packed (adjacent, unpadded) in the feature portion of the buffer?

If so, can we then deduce the precise byte-range of each feature using only the index, by taking the difference between adjacent leaf nodes: `Range: ${leaf_node[i].offset} - ${leaf_node[i+1].offset - 1}`. 

If we can in fact determine the precise byte-range for each feature, we could trivially merge adjacent byte ranges to combine requests. And, like with the new index traversal logic, there's probably a reasonable heuristic for merging non-adjacent ranges, provided they're "close enough" to be worth saving a request.

All of that's predicated on my assumption that we can use the index in this way... what do you think? 

If I'm mistaken, and that's not possible or inordinately complicated, then I support your two proposals.

michaelkirk:
Another thought would be that we could very crudely estimate feature size by taking the `feature_buffer_size / feature count` . By the time we're fetching features, we know how many features we're fetching, so if we want to say... do it in `n` requests, our buffer size could be 
```
// try to get all the feature in 4 requests
let query_count_goal = 4;
let avg_feature_size = (feature_buffer_size / total_feature_count);
feature_count * avg_feature_size / query_count_goal
```

bjornharrtell:
Yes feature data is adjacent and unpadded so your ideas are valid. 🙂

I was thinking an opt in global cache yes. The underlying implementation could also be exposed for more advanced cases but I want to provide a simple as possible API for common use cases.

michaelkirk:
Some more ideas for optimizing index traversal...

[...]

PR #131, opened 2021-06-10
document CDN recommendations for faster spatial filtering

michaelkirk:
This is kind of a strange PR - primarily I'm asking you to click this button in your cloudfront distribution so our examples can run faster:

cache-options

But also, I thought it might make sense to document this somewhere for other folks, since it makes a huge difference.

I noted in https://github.com/flatgeobuf/flatgeobuf/issues/122#issuecomment-844546775 an inordinate delay between the 1st and 2nd requests, greater than all the other requests combined:

(screenshot from safari)
image
Apparently the safari devtools has [a bug that greatly confuses the display of preflight (OPTION) requests](https://bugs.webkit.org/show_bug.cgi?id=213215). 

So let's look at Chrome, which makes it much clearer that we're waiting a long time for preflight OPTIONS request:
Screen Shot 2021-06-10 at 11 24 17 AM

Browsers typically enforce a very short TTL for caching OPTIONS requests, so even subsequent requests still pay the OPTIONS tax (while caching the GETs):
Screen Shot 2021-06-10 at 11 38 58 AM

So we can't rely on the browser to cache it. Why is it so much slower than our other requests?

Screen Shot 2021-06-10 at 1 20 20 PM

Ah, a cache miss from CloudFront! Turns out CloudFront doesn't enable caching options requests by default, but they can [easily be enabled](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/header-caching.html#header-caching-web-cors).

I tested this against my own cloudfront config.

**Before**
Screen Shot 2021-06-10 at 12 54 51 PM

(note that at 298ms the preflight request is already faster in my baseline case, but I suspect that's only because I'm physically nearer to my own s3 origin than I am to the s3 origin used by flatgeobuf.septima.dk).

**After**
Screen Shot 2021-06-10 at 12 59 10 PM
🚀

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,30 @@ As performance is highly data dependent I've also made similar tests on a larger
 | Write w/spatial index | 1         | 1.07       | 0.70       |
 | Size                  | 1         | 0.77       | 0.95       |
 
+### Optimizing Remotely Hosted FlatGeobufs
+
+If you're accessing a FlatGeobuf file over HTTP, consider using a CDN to
+minimize latency.
+
+In particular, when [using the spatial
+filter](https://flatgeobuf.org/examples/leaflet/filtered.html) to get a subset
+of features, multiple requests will be made. Often round-trip latency, rather
+than throughput, is the limiting factor. A caching CDN can be especially
+helpful here.
+
+Fetching a subset of a file over HTTP utilizes Range requests. If the page
+accessing the FGB is hosted on a different domain from the CDN, [Cross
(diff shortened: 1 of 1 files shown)


bjornharrtell (approved):

bjornharrtell:
Nice observation and it all makes good sense. I've enabled the option at CloudFront and it should be propagated.

michaelkirk:
Verified that https://flatgeobuf.org/examples/leaflet/large.html is looking much better for me.

Screen Shot 2021-06-10 at 2 19 58 PM

PR #133, opened 2021-06-15
ts: update to reflect new generated files location

michaelkirk:
The first commit reflects where the files are created when I run `script/gen_ts.sh`

I don't know if this reflects a change in behavior of flatc or some configuration that I'm missing.

The second commit is updating imports to reference the new location.

```
$ flatc --version
flatc version 2.0.0
```


bjornharrtell:
Previously I've temporarily removed the namespace declaration depending on language. But I think since the new modularized output from flatc the namespace is sensible.

PR #134, opened 2021-06-15
ts: fmt + fix lint

michaelkirk:
Based on #133, so you might want to review that first. If you prefer, I can hold off and rebase this once #133 is addressed.

---

This introduces a couple of house-cleaning things...

Potentially controversial: I'm proposing we adopt a code formatter for typescript - we've got several inconsistent styles going on throughout the code base. 

I've used `prettier` since it seems to be the most popular. I care less about *which* particular tool or code style we adopt, so long as we have something automatic.

I've also fixed up eslint and addressed the existing lints.

If the formatting stuff is controversial, I can back that out and apply just the lint fixes.

The diff is large, but it's mostly formatting changes. I'd recommend looking at the individual commits, which are intended to be meaningful.


bjornharrtell:
Not controversial for me. I'm all for merging this and getting prettier in place.

michaelkirk (commented):
I've rebased and pushed. LMK if you'd like any changes.

PR #135, opened 2021-06-16
js: bbox fetches features in parallel batches

michaelkirk:
When fetching features from a bbox query, long ago we used to very
naively fetch the precise range of each feature. This was efficient in
bytes, but resulted in lots of requests.

As of 860c411 we arbitrarily fetch some extra bytes on the chance that it'd obviate a
future request. In practice this performs fairly well - much better
than the naive approach. But you'd always have edge cases where you were
just a few bytes short of needed features, causing extra requests.

Rather than arbitrarily fetching an extra N bytes every time we fetch a
feature and hope it covers some soon-to-be-needed features, if we
instead knew the precise location and lengths of all the features we
need, we could explicitly combine the nearby ones into a single request.

Since the spatial index has the offsets of every feature*, we can do just
that - right-sizing our feature requests and surfacing a configuration
to let people tune the number of bytes they are willing to "spend" in
order to obviate a future request (see [src/ts/Config.ts]).

*the one exception is the very last feature. Since the "next" feature
serves as the bookend of "this* feature, we are left to just guess the
length of the last feature. In practice requesting a Range beyond the
end of the file results in fetching precisely what we need.

## Example

I've also spruced up the example a bit to highlight the feature properties as well as the geometry. Now that the client performs a bit better, I've also allowed the example to cover more map area. 

**after on the left / before on the right**
image

### DIFF
--- a/examples/leaflet/index.html
+++ b/examples/leaflet/index.html
@@ -1,8 +1,8 @@
 <html>
 <head>
-    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.6.0/dist/leaflet.css" />
+    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.7.1/dist/leaflet.css" />
     <link rel="stylesheet" href="/examples/site.css" />
-    <script src="https://unpkg.com/leaflet@1.6.0/dist/leaflet.js"></script>
+    <script src="https://unpkg.com/leaflet@1.7.1/dist/leaflet.js"></script>
     <script src="https://unpkg.com/flatgeobuf@3.17.4/dist/flatgeobuf-geojson.min.js"></script>
     <script src="https://unpkg.com/json-formatter-js"></script>
 
--- a/src/ts/HttpReader.ts
+++ b/src/ts/HttpReader.ts
@@ -1,4 +1,5 @@
 import * as flatbuffers from 'flatbuffers';
+import { Repeater } from '@repeaterjs/repeater';
 
 import {
     Rect,
@@ -8,13 +9,13 @@ import {
     streamSearch,
 } from './packedrtree';
 import { magicbytes, SIZE_PREFIX_LEN } from './constants';
+import Config from './Config';
 import Logger from './Logger';
 import HeaderMeta from './HeaderMeta';
 import { Feature } from './flat-geobuf/feature';
 
 export class HttpReader {
     private headerClient: BufferedHttpRangeClient;
-    private _featureClient?: BufferedHttpRangeClient;
     public header: HeaderMeta;
     private headerLength: number;
     private indexLength: number;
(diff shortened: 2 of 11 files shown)


examples/leaflet/index.html
```diff
@@ -1,8 +1,8 @@
 <html>
 <head>
-    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.6.0/dist/leaflet.css" />
+    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.7.1/dist/leaflet.css" />
```
michaelkirk, line 3: I also updated leaflet in the examples.

src/ts/HttpReader.ts
```diff
@@ -266,13 +340,40 @@ class HttpRangeClient {
...
         const range = `bytes=${begin}-${begin + length - 1}`;
-        Logger.debug(
+        Logger.info(
             `request: #${this.requestsEverMade}, purpose: ${purpose}), bytes: (this_request: ${length}, ever: ${this.bytesEverRequested}), Range: ${range}`
         );
 
         const response = await fetch(this.url, {
             headers: {
                 Range: range,
+                // TODO: better parallelize requests on Chrome
```
michaelkirk, line 350: This is really unfortunate. 

image

Overall I think the changes are still worthwhile, even for chrome.

1. batching reduces number of requests and overall request size
2. Though it's significantly worse than Safari/Chrome, there is still *some* concurrency with Chrome - the second request starts as soon as we have the headers for the first. 

This example shows some longer requests - during which Chrome can benefit from concurrent feature fetching a bit more - notice how the 4th/5th request complete during the time the 3rd larger request is downloading.

image

bjornharrtell: Curious...I wonder if we can expect Chrome to fix this eventually? Seems like a rather big dent in it's performance track record.

michaelkirk: In case you missed it from the comment, this [https://bugs.chromium.org/p/chromium/issues/detail?id=969828](chrome bug), was filed two years ago.

My impression is that Chrome (and probably other folks) primarily think of range requests as being for:

- resuming broken downloads
- streaming/seeking through media

So our use case is pretty niche. The only other people apart from us that I'm aware of using this are bit torrent  implementations (webtorrent). I've also seen parallel Range requests used for privacy purposes (occluding from the outside network what content is being requested).

I don't have any way of knowing, but I don't expect it to be fixed soon.

michaelkirk: A humble plea to "star" the issue if you have a google account: https://bugs.chromium.org/p/chromium/issues/detail?id=969828

bjornharrtell (approved):

src/ts/HttpReader.ts
bjornharrtell: done :)

PR #152, opened 2021-11-04
Rust writer

pka:
WIP - feedback welcome.


src/rust/src/feature_writer.rs
```diff
@@ -22,6 +23,7 @@ pub struct FeatureWriter<'a> {
     parts: Vec<flatbuffers::WIPOffset<Geometry<'a>>>,
     properties: Vec<u8>,
     fbb: flatbuffers::FlatBufferBuilder<'a>,
+    pub(crate) bbox: NodeItem,
```
bjornharrtell, line 26: hmm what does pub(crate) mean?

pka: Makes `bbox` public within the crate only. So it can be accessed in `file_writer.rs`, but it's not visible for users of the library.

bjornharrtell (approved):
LGTM

src/rust/src/properties_reader.rs
```diff
@@ -77,11 +77,15 @@ impl geozero::FeatureProperties for FgbFeature {
         let mut finish = false;
         if let Some(properties) = self.fbs_feature().properties() {
             let mut offset = 0;
-            while offset < properties.len() - 1 && !finish {
+            while offset + 1 < properties.len() && !finish {
                 // NOTE: it should be offset < properties.len(), but there is data with a
                 // trailing byte in the last column of type Binary
                 let i = LittleEndian::read_u16(&properties[offset..offset + 2]) as usize;
                 offset += size_of::<u16>();
+                if i >= columns_meta.len() {
```
michaelkirk, line 85: FGB allows properties to be specified either:

1. defined once, in the header, all features are uniformly assumed to have the same props
2. defined per feature, each feature can have different properties.

Am I correct in understanding that this is only writing the uniform header properties, and that per-feature properties are not currently supported?

pka: I totally forgot about this, TBH. But as far as I see, there is actually no restriction in the writer implementation! The code you're referring to, is part of the existing reader implementation, which needs property type information from the header. For full per-feature property support, we would need an additional reader method with a property type argument.

PR #153, opened 2021-11-10
Use http-range-client crate

pka:
https://crates.io/crates/http-range-client


michaelkirk:
Am I correct that this is strictly an extraction of the existing logic into a new crate?  Or were there changes in the new crate vs. when it was internal?

pka:
Yes, it's exactly the same logic. The only difference of the crate is, that it's prepared for supporting other HTTP clients via feature flags. The reason for extracting http-range-client was to enable reuse for other cloud optimized formats like 3D tiles, COG or COPC.

michaelkirk (approved):

bjornharrtell (approved):

PR #154, opened 2021-11-13
Replace unsafe casts in index reader

pka:


src/rust/src/packed_r_tree.rs
```diff
@@ -45,6 +47,39 @@ impl NodeItem {
         }
     }
 
+    fn from_bytes(raw: &[u8]) -> Result<Self> {
+        if raw.len() < size_of::<NodeItem>() {
+            return Err(GeozeroError::GeometryIndex);
+        }
+        let conv_err = { |_| GeozeroError::GeometryIndex };
+        Ok(NodeItem {
+            min_x: f64::from_le_bytes(raw[0..8].try_into().map_err(conv_err)?),
```
michaelkirk, line 56: Do you think it'd be worth de-duping the offset logic with something like:

```
Self::from_reader(&mut std::io::Cursor::new(raw))
```

src/rust/src/packed_r_tree.rs
```diff
@@ -153,10 +181,8 @@ async fn read_http_node_items(
         .map_err(from_http_err)?;
 
     let mut node_items = Vec::with_capacity(length);
-    let buf = unsafe { std::slice::from_raw_parts_mut(node_items.as_mut_ptr() as *mut u8, len) };
-    buf.clone_from_slice(&bytes);
-    unsafe {
-        node_items.set_len(length);
+    for i in 0..length {
+        node_items.push(NodeItem::from_bytes(&bytes[i * size_of::<NodeItem>()..])?);
```
michaelkirk, line 185: Maybe better to be explicit about this range length rather than ".."  to make sure `from_bytes` isn't reading beyond the desired bounds.

michaelkirk (approved):
Nice work! I'm happy to see some usage of `unsafe` go away. 

I had some small feedback, but none of it seems critical. Feel free to ignore it.

src/rust/src/packed_r_tree.rs
pka: Thanks for the great tip, I always forget about that!

bjornharrtell (approved):

PR #158, opened 2021-11-24
add test for writing fgb and loading it after with FgbWriter and FgbR…

highway900:
Adds a failing test that writes a FGB file using `FgbWriter` and attempts to read the file back in using `FgbReader`.

Writing is successful but reading is not returning an error `IoError(Error { kind: UnexpectedEof, message: "failed to fill whole buffer" })`

I have noted that files created using the `FgbWriter` have resulted in features counts double what they should be, it is possible I am not correctly setting the header `features_count` and the header is corrupt causing the file to load incorrectly.

The file does load with GDAL and shows the doubling of features count. These extra features are filtered out in the attribute editor of QGIS, so they seem to partially exist?

@pka I hope this PR helps. My `Result` handling could do with some work.

### DIFF
--- a/src/rust/tests/write.rs
+++ b/src/rust/tests/write.rs
@@ -6,7 +6,7 @@ use geozero::geojson::{GeoJson, GeoJsonReader};
 use geozero::{ColumnValue, GeozeroDatasource, PropertyProcessor};
 use std::fs::File;
 use std::io::{BufReader, BufWriter, Write};
-use tempfile::tempfile;
+use tempfile::{tempfile, NamedTempFile};
 
 #[test]
 fn write_file() -> std::io::Result<()> {
@@ -141,8 +141,8 @@ fn geozero_to_fgb() -> Result<()> {
 }
 
 #[test]
-fn test_save_fgb_and_load() {
-    let tempf = tempfile().expect("test failed");
+fn test_save_fgb_and_load() -> Result<()> {
+    let file_to_write = NamedTempFile::new()?;
 
     // Save
     let linestrings: Vec<LineString<f64>> = vec![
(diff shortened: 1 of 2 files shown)


src/rust/tests/write.rs
```diff
@@ -138,3 +139,45 @@ fn geozero_to_fgb() -> Result<()> {
...
+            (x: -21.95156, y: 64.1446),
+            (x: -21.951, y: 64.14479),
+            (x: -21.95044, y: 64.14527),
+            (x: -21.951445, y: 64.145508)
+        ],
+        geo_types::line_string![(x: 0.0, y: 0.0), (x: 1.0, y: 1.0),],
+    ];
+
+    let mut fgb = FgbWriter::create("test_write", GeometryType::LineString, |_fbb, header| {
+        header.features_count = linestrings.len() as u64;
```
pka, line 159: This line causes ` failed to fill whole buffer` when reading, because `features_count` is incremented with `add_feature_geom`

src/rust/tests/write.rs
```diff
@@ -138,3 +139,45 @@ fn geozero_to_fgb() -> Result<()> {
...
+
+    for geom in linestrings.iter() {
+        let geom: geo_types::Geometry<f64> = geom.to_owned().into();
+        fgb.add_feature_geom(geom, |_feat| {}).expect("test failed");
+    }
+    let mut file = BufWriter::new(&tempf);
+    fgb.write(&mut file).expect("test failed");
+
+    // Load
+    let mut filein = BufReader::new(&tempf);
```
pka, line 173: Reading the just written tempfile seems not to work. It's working with a regular file and an additional `file.flush()`.

pka (changes requested):

src/rust/tests/write.rs
highway900: I have removed the explicit feature count setting.

src/rust/tests/write.rs
highway900: Using `NamedTempFile` now and can call `reopen` after `flush` to continue using temp file for read. Test passes.

highway900:
Updated PR to include suggested fixes. Test passes.

pka (approved):
Thanks, great usage example!

highway900:
No worries! Glad I can help.

PR #159, opened 2021-11-24
Handle heterogeneous geometry types in TS impl

worace:
In some experiments I was doing, I noticed the current TS implementation only relies on the specified GeometryType in the header, which means it can't handle files containing heterogeneous geometry types where the header says GeometryType: Unknown


src/ts/generic/geometry.ts
```diff
@@ -80,26 +81,23 @@ export function flat(
     }
 }
 
-export function parseGeometry(
-    geometry: ISimpleGeometry,
-    type: GeometryType
```
worace, line 85: So at the moment I simplified this to never accept the provided geometry type, and always rely on the geometry type specified in the feature. I need to research whether this is problematic or not.

It raises the question of whether there are cases where people would prefer to set the type in the Header only and not in the individual features (e.g. if your file has only points, set point in the header and leave type blank on the individual geometries). I need to confirm whether that's actually valid in the format, and if so whether there are any compelling reasons for supporting it.

bjornharrtell: Yes the "normal" is that homogenous collections only have defined type and columns in the header. Default values in Flatbuffers do not need any storage space so there is a good reason for it too. 🙂

worace: That makes sense. So in that case we need to still probably support this parameter (rather than removing it as I have done), but the logic would be that if the provided GeometryType is "unknown", then we look at the type of the individual geometry.

bjornharrtell: Yep, I want this done so I've made https://github.com/flatgeobuf/flatgeobuf/issues/161 for it.

worace: Sounds good. I updated this branch with the change discussed above. I saw you merged the other PR with the formatter update so in the morning I will rebase this one and then it should be good to go I think.

bjornharrtell: Ok I'll try to get time to review this. I think it at least needs some test cases though.

worace: Sounds good. I had added a couple but I think they were hard to see due to the big formatter diff in that spec file. But I added a few more now as well, let me know what you think.

worace: Also I see the CI is running now, thanks :+1:

src/ts/generic/featurecollection.ts
```diff
@@ -192,10 +193,30 @@ function valueToType(value: boolean | number | string): ColumnType {
 }
 
 function introspectHeaderMeta(features: IFeature[]): HeaderMeta {
-    const feature = features[0];
-    const geometry = feature.getGeometry ? feature.getGeometry() : undefined;
-    const geometryType = geometry ? geometry.getType() : undefined;
-    const properties = feature.getProperties ? feature.getProperties() : {};
+    let geometryType: GeometryType = undefined;
+
+    for (const f of features) {
```
worace, line 198: 1 thing I haven't really wrapped my head around is why there are 2 implementations of many of these functions, 1 under `generic/` and 1 under `geojson/`. I duplicated this logic in both places at the moment which obviously isn't ideal.

We could potentially pull it out into some helper, but I don't think the `IFeature` and `IGeoJsonFeature` types are interchangeable. So it might need to be something like `inferGeometryType(geomTypes: string[])`. Curious if anyone has thoughts or knows more of the backstory here

bjornharrtell: I've tried to reduce the duplication in that generic contains most of the logic but as you note I've not been entirely successful. The two implementation targets geosjon and ol (OpenLayers) uses generic. IFeature is the generic interface and is inherited in IGeoJsonFeature and ISimpleFeature (OpenLayers) which are quite different in structure.

worace: Maybe something like this? https://github.com/flatgeobuf/flatgeobuf/pull/159/commits/60f952b140064135cd21015d5abb168138c57a29

It's a little clumsy b/c even tho `IGeoJsonFeature` extends `IFeature`, it lacks the `getGeometry` interface we would need to treat them consistently. I'm not sure if there's a more Typescipt-idiomatic way to do the type of disambiguation I'm doing in that function.

No rush on reviewing, I will probably be a way for a bit for Thanksgiving Holiday. :turkey:

bjornharrtell: Had a look and I like what I see. :) Thanks.

bjornharrtell (approved):
LGTM. Shall I go ahead and merge @worace?

worace:
Sounds good! Thanks!

bjornharrtell:
Thanks for the contribution @worace, much appreciated!

PR #165, opened 2021-12-04
Run tsc compiler in CI

worace:
Would it be reasonable to run this in CI as a guard against mistakes like I made in https://github.com/flatgeobuf/flatgeobuf/commit/3b4e22c33aac74cbb96ec6c6b299c20a692d8948 ?


bjornharrtell (approved):

bjornharrtell:
Yep, thanks. 🙂

PR #172, opened 2022-02-09
Handle empty columns in header in rust reader

worace:
Here's an issue I ran into comparing between some of the different implementations -- the rust reader seems to reject FGB with empty columns entries in the header, even tho the field is supposedly marked optional in the spec. The JS reader seems to handle these files OK.

Unfortunately, I've had a tough time testing this, because the rust writer seems to always populate this field with a  `Some` value even if there are no columns/properties.

The test I included in this PR is an example of what I tried to use to test this, but this test doesn't actually work -- it passes even before the fix proposed in this PR.

I suppose I could add a statically-built .fgb test file in the repo for this case. Curious if anyone has thoughts on this.


src/rust/tests/write.rs
```diff
@@ -129,6 +129,35 @@ fn json_to_fgb() -> Result<()> {
...
 }
 
+#[test]
+fn json_to_fgb_geometry_only() -> Result<()> {
+    use std::io::Cursor;
+
+    let mut fgb = FgbWriter::create("countries", GeometryType::Point, |fbb, header_args| {
+        header_args.description = Some(fbb.create_string("Point no properties"));
+        header_args.columns = None;
+    })?;
```
worace, line 139: Here's some debug output showing the header that the rust writer currently generates for this. Note the `columns: Some([])` even tho I manually set it to None in the args.

```
[src/properties_reader.rs:48] self.header() = Header {
    name: Some(
        "test_point",
    ),
    envelope: Some(
        [
            0.0,
            0.0,
            0.0,
            0.0,
        ],
    ),
    geometry_type: Point,
    hasZ: false,
    hasM: false,
    hasT: false,
    hasTM: false,
    columns: Some(
        [],
    ),
    features_count: 1,
    index_node_size: 16,
    crs: Some(
        Crs {
            org: None,
            code: 4326,
            name: None,
            description: None,
            wkt: None,
            code_string: None,
        },
    ),
    title: None,
    description: Some(
        "Point no properties",
    ),
    metadata: None,
}
```

bjornharrtell:
Seems like a flaw in the Rust implementation to me. Or is this general behaviour of the Rust implementation? That would not be good especially considering that the point of unset optional/default fields is that they take up no wire space. @pka thoughts?

pka:
According to the comment, this should have been fixed in a85f934. Maybe it's already failing in https://github.com/flatgeobuf/flatgeobuf/blob/cefb58f19842e5e72f85d81bf16cc43b0dd208e7/src/rust/src/properties_reader.rs#L51 ?

pka:
Your fix looks good! And yes, we probably need a static test file. 

The writer replaces `header_args.columns` with the content of `self.columns`, which is an array and can only be changed by calling `add_column`. So it's currently only possible to write an empty array of columns.

pka:
> Seems like a flaw in the Rust implementation to me. Or is this general behaviour of the Rust implementation? That would not be good especially considering that the point of unset optional/default fields is that they take up no wire space. @pka thoughts?

This behaviour should be specific to the columns header entry. Other fields have the Flatbuffers default value.

bjornharrtell (approved):

bjornharrtell:
So this is good to merge @pka?

bjornharrtell:
Thanks @worace :)

PR #173, opened 2022-02-17
Handle Json Column Type in typescript impl

worace:
Is this a correct read of the [JSON](https://github.com/flatgeobuf/flatgeobuf/blob/master/src/fbs/header.fbs#L39) column type?

I interpreted it to mean another UTF-8 string but with the assumption that it contains valid encoded JSON, and therefore could either be used to encode more complex JSON types (like objects or arrays), or potentially as an escape hatch for heterogeneous datasets.

### DIFF
--- a/src/ts/geojson/featurecollection.ts
+++ b/src/ts/geojson/featurecollection.ts
@@ -82,7 +82,8 @@ function valueToType(value: boolean | number | string | unknown): ColumnType {
         else return ColumnType.Double;
     else if (typeof value === 'string') return ColumnType.String;
     else if (value === null) return ColumnType.String;
-    else return ColumnType.Json;
+    else if (typeof value === 'object') return ColumnType.Json;
+    else throw new Error(`Unknown type (value '${value}')`);
 }
 
 function introspectHeaderMeta(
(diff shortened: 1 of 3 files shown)


bjornharrtell:
@worace yep that is the general idea of the JSON type.

worace:
@bjornharrtell do you think this is a reasonable way to handle that type in JS/TS? That is, auto-parse fields declared as JSON when deserializing, and use it as the default serialization type for non-primitive JS types like objects or arrays etc?

The downside I could imagine is some people might want more tight control over how those get handled or when they get parsed. But as it is currently the TS impl doesn't handle them at all, so it would be nice to be able to extract them somehow.

src/ts/geojson/featurecollection.ts
```diff
@@ -82,7 +82,7 @@ function valueToType(value: boolean | number | string | unknown): ColumnType {
         else return ColumnType.Double;
     else if (typeof value === 'string') return ColumnType.String;
     else if (value === null) return ColumnType.String;
-    else throw new Error(`Unknown type (value '${value}')`);
+    else return ColumnType.Json;
```
bjornharrtell, line 85: perhaps be more explicit here with else if (typeof value === 'object')?

jxeeno:
Hey, thanks for working on this @worace! I'm keen to get JSON types added for a project. Do you have some time to update the type check? Else I'm happy to jump in :)

worace:
Oops sorry guys i got distracted and let this sit for so long. Just made that tweak. Let me know what you think.

bjornharrtell:
LGTM, thanks. :)

PR #174, opened 2022-02-21
Attempt to allow different output formats

bjornharrtell:
~~While this compiles I get empty output with --input ../../../test/data/countries.fgb > ../../../test/data/countries2.geojson.~~

~~Don't understand what is causing that but suspect is an important puzzle for me to solve to understand something fundamental in Rust.. (perhaps related to the lifetime specifier? or the generic W parameter for the GeoJsonWriter type that is confusing me?~~

Update: the above had to with use of stdout.. with files it works.

Now I have another question.. I'd like to understand how to break about the main into methods but the outcommented atempt doesn not work because "GeozeroDatasource" is not object safe. I would like to know more about this and how I can break out the code...


src/rust/fgbutil/src/main.rs
```diff
@@ -34,13 +38,54 @@ enum Format {
     Geojson,
 }
 
+// TODO: Cannot pass around GeozeroDatasource because it is a non object safe trait?
+// TODO: Regardless might need dyn / Box?
+
+fn write(format: Format, reader: GeozeroDatasource, output: BufWriter<File>) -> Result<()> {
```
michaelkirk, line 44: > Now I have another question.. I'd like to understand how to break about the main into methods but the outcommented atempt doesn not work because "GeozeroDatasource" is not object safe. I would like to know more about this and how I can break out the code...

tldr;
```
fn write(format: Format, reader: impl GeozeroDatasource, output: BufWriter<File>) -> Result<()> {
```

Longer answer (someone please correct or add to this if I'm wrong):

As you probably know, GeozeroDatasource is a trait. It can't be instantiated on its own. Instead, it's behavior that can be ascribed to potentially many different instantiable things. Those things are said to "implement GeozeroDatasource".

A "trait objects" is a pointer that could point to any thing that implements the trait. It's spelled like `Box<dyn GeozeroDatasource>`. 

As I understand it, this format is similar to how most other languages pass around trait-like things (e.g. arguments as protocols in objective-c/swift, or interfaces in java).

There are implications to this, rust errs on the side of being explicit, which is why it's such a mouthful to spell in rust. 

So in situations like this you *could* use a trait object, but only "object safe" traits can be made into a trait object. 

But what does "object safe" mean?
See https://doc.rust-lang.org/book/ch17-02-trait-objects.html#object-safety-is-required-for-trait-objects for the gospel.

But, since I know you write some java, a more intuitive example might be to read this persons question: https://stackoverflow.com/questions/8161712/how-can-i-refer-to-the-class-type-a-interface-is-implementing-in-java

I think that's philosophically similar issue - there are some methods defined in a trait that would need to *know* what the underlying implementing type is.

Since "trait objects" aren't usable in this case, another option we have to say "this argument could be one of  potentially different types that implement `GeoZeroDatasource`", is to use generics to compile a version of `write` for every underlying type that ultimately uses it.

```
fn write<G: GeozeroDatasource >(format: Format, reader: G, output: BufWriter<File>) -> Result<()> {
```

There is the "impl trait" sugar to equivalently spell this:

```
fn write(format: Format, reader: impl GeozeroDatasource, output: BufWriter<File>) -> Result<()> {
```

When both trait objects or generics are an option, there are pros and cons to either. (primarily: generics can perform better, but because generics produce multiple versions of the function,  it's slower to compile and bloats code size).

src/rust/fgbutil/src/main.rs
```diff
@@ -34,13 +38,54 @@ enum Format {
...
+    match format {
+        Format::Geojson => write_flatgeobuf(reader, output)?,
+        Format::Flatgeobuf => write_flatgeobuf(reader, output)?
+    }
+    Ok(())
+}
+
+fn write_geojson(reader: GeozeroDatasource, output: BufWriter<File>) -> Result<()> {
```
pka, line 52: `reader: GeozeroDatasource` should be `mut reader: impl GeozeroDatasource`

pka (changes requested):
Compiling code:
```
fn write(format: Format, reader: impl GeozeroDatasource, output: BufWriter<File>) -> Result<()> {
    match format {
        Format::Geojson => write_flatgeobuf(reader, output)?,
        Format::Flatgeobuf => write_flatgeobuf(reader, output)?
    }
    Ok(())
}

fn write_geojson(mut reader: impl GeozeroDatasource, mut output: BufWriter<File>) -> Result<()> {
    let mut writer = GeoJsonWriter::new(&mut output);
    reader.process(&mut writer)?;
    Ok(())
}

fn write_flatgeobuf(mut reader: impl GeozeroDatasource, mut output: BufWriter<File>) -> Result<()> {
    // TODO: would make sense if GeozeroDatasource could provide name and geometry_type?
    let name = "";
    let geometry_type = flatgeobuf::GeometryType::Unknown;
    let mut writer = FgbWriter::create(name, geometry_type, |_, _| {})?;
    reader.process(&mut writer)?;
    writer.write(&mut output)?;
    Ok(())
}

fn transform(inputformat: Format, outputformat: Format, mut input: BufReader<File>, output: BufWriter<File>) -> Result<()> {
    match inputformat {
        Format::Geojson => {
            // TODO: impl..
        },
        Format::Flatgeobuf => {
            let reader = FgbReader::open(&mut input)?;
            let reader = reader.select_all()?;
            write(outputformat, reader, output);
        }
    }
    Ok(())
}
```

src/rust/fgbutil/src/main.rs
```diff
@@ -9,21 +11,23 @@ use clap::{ArgEnum, Parser};
 #[derive(Parser)]
 #[clap(author, version, about, long_about = None)]
 struct Args {
-    /// Input path
```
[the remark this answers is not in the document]
bjornharrtell: Ah, nice - wasn't aware 3 slashes had meaning.

src/rust/fgbutil/src/main.rs
bjornharrtell: Thanks @michaelkirk, it makes sense and explains why the changes suggested by @pka works. I still have issues internalizing the trait/impl part of the Rust type system but at least this is a step in the right direction. :)

src/rust/fgbutil/src/main.rs
bjornharrtell: Thanks @pka, it does work and I now know a little more Rust. :)

Issue #167, opened 2022-01-03
Error (panicked) at FgbReader::open when reading an EPSG:25832 fgb file


tumluliu:
Hi there, thank you for making such a great lib! I'm pretty new to both rust and flatgeobuf, and facing an issue when opening a fgb dataset with CRS EPSG:25832. It complains with:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Geometry("Utf8 error for string in 196..1969: invalid utf-8 sequence of 1 bytes from index 1542\n\twhile verifying table field `wkt` at position 188\n\twhile verifying table field `crs` at position 60\n")'
```

The dataset is a subset of Bavaria cycling network from https://www.ldbv.bayern.de/produkte/weitere/opendata.html , selected and exported to fgb format with QGIS. I am able to open it in QGIS, but met the above issue when reading it with flatgeobuf lib. For other dataset with EPSG:4326 system, there is no such problem. But I am not sure if the CRS is the reason. Could you help shed some light? Thanks a lot! For some reason, I could not upload the sample dataset here. sorry about that!

bjornharrtell:
Thanks for your interest @tumluliu, this does seem unexpected and I will try to get to the bottom if it. Likely need help from Rust maintainer @pka though.

pka:
Hi @tumluliu

I've converted the Shapfile to FGB with GDAL 3.4.0:
```
ogr2ogr -f FlatGeobuf Radweg_gesamt.fgb  Radweg_gesamt.shp
```

ogrinfo -al -so Radweg_gesamt.fgb
```
INFO: Open of `Radweg_gesamt.fgb'
      using driver `FlatGeobuf' successful.

Layer name: Radweg_gesamt
Geometry: Line String
Feature Count: 34974
Extent: (495737.725000, 5238334.501000) - (871457.835000, 5604729.181000)
Layer SRS WKT:
PROJCRS["ETRS89 / UTM zone 32N",
    BASEGEOGCRS["ETRS89",
        DATUM["European Terrestrial Reference System 1989",
            ELLIPSOID["GRS 1980",6378137,298.257222101,
                LENGTHUNIT["metre",1]]],
        PRIMEM["Greenwich",0,
            ANGLEUNIT["degree",0.0174532925199433]],
        ID["EPSG",4258]],
    CONVERSION["UTM zone 32N",
        METHOD["Transverse Mercator",
            ID["EPSG",9807]],
[...]

PR #175, opened 2022-02-22
Control state of FgbReader and Writer / Make FB verification optional

pka:
Solves #167 and avoids wrong use of reader and writer API.

It's a breaking change for the reader API.


bjornharrtell (approved):
Not understanding it fully but looks clean

src/rust/src/http_reader.rs
```diff
@@ -34,8 +44,8 @@ pub(crate) fn from_http_err(error: HttpError) -> GeozeroError {
     }
 }
 
-impl HttpFgbReader {
-    pub async fn open(url: &str) -> Result<HttpFgbReader> {
+impl HttpFgbReader<Initial> {
```
michaelkirk, line 47: I've never quite seen this pattern before! It's a nice way to move run-time checks to compile time without having a bunch of boilerplate types like:

```
struct HttpFgbReaderInner {
     client: BufferedHttpRangeClient,
     /// Current read offset
     pos: usize,
 @@ -25,6 +27,14 @@ pub struct HttpFgbReader {
     item_filter: Option<Vec<packed_r_tree::SearchResultItem>>,
     /// Current position in item_filter
     feat_no: usize,
 }

struct InitialHttpFgbReader(HttpFgbReaderInner);
struct OpenedHttpFgbReader(HttpFgbReaderInner);
struct FeatureSelectedHttpFgbReader(HttpFgbReaderInner);

impl InitialHttpFgbReader {
    pub async fn open(url: &str) -> Result<OpenHttpFgbReader> {
        ...
    }
}
```

Nice work!

pka: Quite proud of that commit :smile:  But credit goes to https://nostarch.com/rust-rustaceans

Issue #178, opened 2022-03-03
Improve GeoJSON Typescript definitions


drzax:
Using the GeoJSON JS library in a typescript project I'm finding a lot of work is required to ensure the declared return types (`IGeoJsonFeatureCollection`, `IGeoJsonFeature`, `IGeoJsonGeometry`, etc) can be reconciled with types expected by functions/libraries that have more strictly correct definitions for GeoJSON objects.

For example, using the Turf.js helper functions with data returned from flatgeobuf requires either type assertions (`return res as FeatureCollection;`) or a bunch of unnecessary type guard runtime code.

Turf actually went through their own migration to use the DefinitelyTyped GeoJSON definitions: https://github.com/Turfjs/turf/pull/2158

Is there any appetite for adopting those definitions here?

bjornharrtell:
@drzax yeah this seems like a good idea, thanks for the pointer. :)

PR #179, opened 2022-03-06
Rework to use @types/geojson

bjornharrtell:
Intended to resolve https://github.com/flatgeobuf/flatgeobuf/issues/178.

It actually made the implementation more compliant, as it seems the GeoJSON RFC and types make the feature properties member mandatory. I'm not sure I agree with it but standard is standard.

@drzax can you review this and even better, try it out to see if it solves your use case?


drzax:
@bjornharrtell — thanks for looking into it so quickly! At first glance this looks good, but not sure I'll find the time to look closely at the code this week. But I will have a go putting it into my project.

Issue #180, opened 2022-03-09
NET: Implementation of CoordinateSequence for FlatGeobuf needs proper Copy method


FObermaier:
The current implementation of `CoordinateSequence` for FlatGeobuf needs a proper `CoordinateSequence.Copy` method. Without it, calls to `Geometry.Copy()` return empty geometries.

FObermaier:
@bjornharrtell, I'm sorry but the initial implementation is buggy for multi-part geometries due to wrong offset for System.Buffers.Copy(...).

bjornharrtell:
@FObermaier does it make sense to reopen this or treat as new bug?

FObermaier:
I think that this will do

PR #181, opened 2022-03-09
Implement FlatGeobufCoordinateSequence.Copy properly.

FObermaier:
Returning null in this method causes Geometry.Copy() to fail.

relates to #180


bjornharrtell:
Looks great. thank you @FObermaier!

bjornharrtell (approved):

Issue #182, opened 2022-03-12
RangeError: Offset is outside the bounds of the DataView


jxeeno:
**Version:** 3.22.0

The TS implementation of flatgeobuf has a number of memory allocation issues in the `buildFeature` method that can cause the `geojson.serialize` method to crash when feature properties are large.

Take this as an example:

```js
import { geojson } from 'flatgeobuf';
import fs from 'fs';

const features = {
    type: "FeatureCollection",
    features: [
[...]

PR #183, opened 2022-03-12
fix(ts): mem allocation for large feature props

jxeeno:
This PR resolves a number of memory allocation issues in the TS implementation:

* offset being redefined, but setters assume offset is still 0 - as described here: https://github.com/flatgeobuf/flatgeobuf/issues/182
* correctly allocate memory when the `size` value passed to `prep(size)` exceeds `capacity * 2`
* call `prep()` when setting the index
* adds tests to validate against above issues


bjornharrtell (approved):
Nice catch, thanks!

PR #189, opened 2022-03-19
Add sequential read only reader to allow piping from stdin

bjornharrtell:
Initial attempt tried to create variants FgbReader that only implements Read but I couldn't figure out how to do that without clashing with Read + Seek and/or similar code duplication.

So I ended up with a specialized class instead... which results in quite a bit of code duplication. Any thoughts about this @pka?


pka (approved):
Didn't check all tweaks you made, but looks reasonable.
As you say, code duplication is the main problem. We already have a lot of duplication between FileReader and HttpReader, but there is it even more difficult to share code because of the async API.
But I'm fine with merging. Maybe we find a more elegant way later.

pka:
`cargo fmt` tweaks a few lines...

PR #190, opened 2022-03-25
Introduce validity metadata at header level

bjornharrtell:


bdon:
What about `winding` being an enum of `CCW`, `CW` or null?

Some source formats like Shapefile have exterior rings CW, and others like GeoJSON define exterior as CCW. If the metadata indicates CW winding but my reader wants CCW winding, I can just read each polygon ring "backwards" without having to make a copy or do any winding checks. This would preserve the lossless nature of transformation from other well-formed formats independent of how they prefer winding order. 

So the language for `winding` can be like this: if `winding` is CCW, the first ring of Polygon coordinates must be CCW and all rings after CW. If `winding` is CW, the first ring is CW and all rings after CCW. It should recommend that the first ring is the "Outer" one and all ones after are "Inner" but that constraint that would belong to Validity and not Winding.

bjornharrtell:
@bdon I agree and it could make sense to have both winding and validity as enums. 

Enum cannot be null but default is 0 and that could (should) be unknown. See http://google.github.io/flatbuffers/flatbuffers_guide_writing_schema.html.

For valitity one could complicate it further. There are a couple of well known validity definitions in the wild (on the top of my head, OGC, ESRI and SQL Server all have their own definitions). A geometry can be both OGC and ESRI valid. Perhaps the bit_flags of flatbuffer schema can be used to represent the combination of states here. Might seem to go into details here but in my experience it's not uncommon that one has to deal with these definitions.

And furthermore OGC have a definition if a geometry is simple or not (see http://www.tsusiatsoftware.net/jts/javadoc/com/vividsolutions/jts/operation/IsSimpleOp.html). I'm not sure if that is of value. I will probably ignore it for now unless someone brings it up.

bjornharrtell:
@bdon as for winding and ring order I agree that strictly interpreted the order belongs to validity.. but in practice, is there a real use case for representing the known winding without also constraining the ring order? I would think that is academic only.. and as such I would word the ring order as a requirement rather than a recommendation for the known windings.

bdon:
No, I didn't have a specific counterexample in mind; indeed just trying to make the language more airtight about winding not implying validity.

bjornharrtell:
Now reworked to these two enums:

https://github.com/flatgeobuf/flatgeobuf/blob/88cacfade9601d71f520fd8b105edb2c84213e1e/src/fbs/header.fbs#L67-L78

bdon:
Winding lgtm, not an expert on Valid in those situations but did find these links to familiarize myself:

OGC Simple & Valid: https://postgis.net/docs/using_postgis_dbmanagement.html#OGC_Validity
ESRI: https://www.esri.com/arcgis-blog/products/data-reviewer/data-management/invalid-geometry-check-explained/

bjornharrtell:
Closing due to inactivity.

bdon:
LGTM, I don't think this is necessary anymore.

PR #195, opened 2022-04-06
Attempt to use geozero from git

bjornharrtell:


michaelkirk:
To reference unpublished code for the local workspace,wWhat you have works, but typically I'd instead solve this by adding an entry to the patches section, like this:

https://github.com/flatgeobuf/flatgeobuf/compare/master...michaelkirk:mkirk/patch-fgb-dep?expand=1

some advantages being:
1. it's localized to a well known place (you don't have to make and unmake this change in multiple places)
2. it applies transitively to all the crates in the workspace
3. (not relevant here) it also applies transitively to any dependencies

bjornharrtell:
Replaced by https://github.com/flatgeobuf/flatgeobuf/commit/6bf6d7fcfa30add6da439a0387d1835e60c93fb6

PR #196, opened 2022-04-07
FIX `--index` arg for fgbutil

michaelkirk:
Previously it wasn't being used


bjornharrtell:
@michaelkirk thanks ;)

bjornharrtell (approved):

PR #198, opened 2022-04-21
Do not consider unverified open as unsafe rust

bjornharrtell:
I don't think opening trusted data with no verification should be forced to unsafe Rust land.


bjornharrtell:
hehe looks like size_prefixed_root_as_header_unchecked is unsafe itself.. I will find out why! :) Until then, closing..

bjornharrtell:
Ah.. I think that unchecked thing is not about buffer verification at all, I think it's about bypassing some Rust safety for speed.

PR #199, opened 2022-04-21
Rust: Fix reading FGB with unknown feature count

pka:


bjornharrtell (approved):

src/rust/src/file_reader.rs
```diff
@@ -94,21 +97,35 @@ impl<'a, R: Read + Seek> FgbReader<'a, R, Open> {
     /// Select all features.
     pub fn select_all(self) -> Result<FgbReader<'a, R, FeaturesSelected>> {
         let header = self.fbs.header();
-        let count = header.features_count() as usize;
-        let index_size = if header.index_node_size() > 0 {
-            PackedRTree::index_size(count, header.index_node_size())
+        let count = {
```
michaelkirk, line 100: Is there a reason it might be important to distinguish between an FGB with an unknown feature count and an FGB which legitimately has zero features in it?

bjornharrtell: I did think about that and remember that I thought a check for data after the header is good enough.

pka: Right, I'll add a file size check to distinguish between empty datasets and datasets with undefined feature count. Currently `features_count` returns `None` for empty datasets.

PR #203, opened 2022-04-23
Add MapLibre examples

carderne:
**NB** I haven't yet made an example for `large.html`, but will do so shortly.

Just wanted to get this up to check that it's worthwhile.

Used MapLibre (rather than Mapbox) purely so that no token would need to be tied to the examples.


bjornharrtell (approved):
Looks good, thanks! :)

PR #217, opened 2022-05-23
add interactive instruments to the list of supporters

cportele:
cc @ogcscotts


bjornharrtell (approved):
Thanks for the support @cportele !

Issue #218, opened 2022-06-25
Rust - builds with `default-features = false` fail


msalib:
Trying to build without `http-range-client` fails since log imports are missing without it:
```
  Compiling flatgeobuf v0.8.0
error: cannot find macro `info` in this scope
   --> /home/msalib/.cargo/registry/src/github.com-1ecc6299db9ec823/flatgeobuf-0.8.0/src/file_writer.rs:370:17
    |
370 |                 info!(
    |                 ^^^^
    |
    = note: consider importing this macro:
            log::info

error: cannot find macro `info` in this scope
   --> /home/msalib/.cargo/registry/src/github.com-1ecc6299db9ec823/flatgeobuf-0.8.0/src/file_writer.rs:376:17
    |
376 |                 info!(
    |                 ^^^^
    |
    = note: consider importing this macro:
            log::info

```

PR #219, opened 2022-06-25
fix rust no default-features builds

msalib:
See https://github.com/flatgeobuf/flatgeobuf/issues/218


pka:
Thanks @msalib for the fix! I think, removing line 68 in lib.rs would be even better.

@bjornharrtell I let you merge, since you couldn't like the additional CI test run...

bjornharrtell (approved):

PR #225, opened 2022-07-29
Rust: Upgrade deps

bjornharrtell:


bjornharrtell:
@pka does it make sense to target edition 2021?

pka:
> @pka does it make sense to target edition 2021?

I don't see a major advantage right now. Sometimes users ask for building with an old rustc. Especially Debian users with their packaged version, which is currently 1.48.0. So I usually wait with updating rustc requirements as long as possible.

bjornharrtell:
Ok down to 2018 so deps upgrade must be the cause of the test failures

bjornharrtell:
Hmm.. looks like the test failures are due to something else entirely. They seem to have started with CI run of https://github.com/flatgeobuf/flatgeobuf/commit/b08fb2c6a6a11b2b438dbbc44c416fb5e27ce5a4 which has nothing to do with Rust.

bjornharrtell:
Looks like behaviour of geozero feature write has changed to not mutate geometry collections, which make sense.. and hinted by comment https://github.com/flatgeobuf/flatgeobuf/blob/6bf6d7fcfa30add6da439a0387d1835e60c93fb6/src/rust/src/feature_writer.rs#L730. @pka can you confirm?

bjornharrtell:
Looks like https://github.com/flatgeobuf/flatgeobuf/pull/226 did the trick. Rebased.

PR #232, opened 2022-08-15
meaningless reversal

Paradox98:
There is no point in reversing the array twice before and after.

### DIFF
--- a/src/cpp/packedrtree.cpp
+++ b/src/cpp/packedrtree.cpp
@@ -220,12 +220,9 @@ std::vector<std::pair<uint64_t, uint64_t>> PackedRTree::generateLevelBounds(cons
     n = numNodes;
     for (auto size : levelNumNodes)
         levelOffsets.push_back(n -= size);
-    std::reverse(levelOffsets.begin(), levelOffsets.end());
-    std::reverse(levelNumNodes.begin(), levelNumNodes.end());
     std::vector<std::pair<uint64_t, uint64_t>> levelBounds;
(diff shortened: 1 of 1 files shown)


bjornharrtell:
Hmm I will have to think about this. As I recall it was done to turn the index up side down to make it sequentially searchable.

Paradox98:
negative to positive.Please let me know when you have a conclusion.

bjornharrtell:
I don't see any double reversal. It's three different arrays (levelOffsets, levelNumNodes and levelBounds).

Paradox98:
After the two arrays（levelOffsets,levelNumNodes) are reversed, the value is taken in the forward direction, stored in levelBounds, and then reverse the levelBounds array.
is equal, 
the two arrays（levelOffsets,levelNumNodes) not reversed, the value is taken in the forward direction, stored in levelBounds.
so,
It is not necessary to reverse the three times before and after.

Paradox98:
The two arrays（levelOffsets,levelNumNodes） are useless, the only thing returned is levelBounds

Paradox98:
1，2，3
3，2，1

3，2，1
1，2，3
:)

bjornharrtell:
Ok I'll give it some more thought when my brain is more awake. :D

bjornharrtell:
Squash merged in another PR as https://github.com/flatgeobuf/flatgeobuf/commit/342d0efe49f237247e866f0bc261f0f821bc6e6f. Thanks @Paradox98!

PR #233, opened 2022-08-16
remove redundant reversals from all clients

michaelkirk:
This is a follow up to #232, but for all the other clients as well.

Previously we reversed a vec, which was made by (effectively) zipping together two reversed vecs, now we just skip all the reversals for the same end result.

Note that I haven't personally tested the `.cs` changes because I don't have a build environment set up for that, but I'm hoping CI will catch any issues there.

### DIFF
--- a/src/cpp/packedrtree.cpp
+++ b/src/cpp/packedrtree.cpp
@@ -220,12 +220,9 @@ std::vector<std::pair<uint64_t, uint64_t>> PackedRTree::generateLevelBounds(cons
     n = numNodes;
     for (auto size : levelNumNodes)
         levelOffsets.push_back(n -= size);
-    std::reverse(levelOffsets.begin(), levelOffsets.end());
-    std::reverse(levelNumNodes.begin(), levelNumNodes.end());
     std::vector<std::pair<uint64_t, uint64_t>> levelBounds;
     for (size_t i = 0; i < levelNumNodes.size(); i++)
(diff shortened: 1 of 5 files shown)


bjornharrtell:
Ok since you are now two claiming that it's redundant perhaps you are right. :) I still want to think it through some more... and make *absolute* certain it doesn't affect index structure because that would break compatibility. As mentioned elsewhere this logic was changed/added as part of spatial index structure changes in spec v3 with the most important change that the tree was turned up side down. That work was done in https://github.com/flatgeobuf/flatgeobuf/pull/40 and specifically https://github.com/flatgeobuf/flatgeobuf/pull/40/commits/8e9820a30ba857b747038021691afd839c00dfc1 added the reversals.

michaelkirk:
I've added #234 to help verify correctness. With #234 you can easily generate FGB files from geojson to compare outputs.

For example you can run this command on different shas to compare the output of FGB's generated from test/fixtures/*

```sh
cd src/cpp
DIR=output-$(git rev-parse --short HEAD); mkdir -p $DIR; for file in test/fixtures/*; do build/geojson2fgb $file > $DIR/$(basename $file.fgb); done
```

Comparing the output before and after 7d11892700c56488e5a32404df494f11749d83dc produces the same result.

Intentionally breaking things, e.g. removing 2 of the three reversals, generates different binaries.

bjornharrtell:
Hmm same CI issue as in https://github.com/flatgeobuf/flatgeobuf/pull/234#issuecomment-1219753943.

bjornharrtell:
Ok I'm pretty convinced at this point.

PR #234, opened 2022-08-17
standalone cpp example and build files

### LINKED ISSUE #194: build scripts and docs for cpp
I wanted to compare some behavior of the rust writer implementation to the cpp implementation, but I was unable to find any documentation or configuration to build the cpp implementation. 

I know that there is build support for the version [embedded in GDAL](https://github.com/OSGeo/gdal/blob/master/ogr/ogrsf_frmts/flatgeobuf), but it seems desirable that it can be built and tested in a stand alone way. 

What do you think? Is this something that you have locally @bjornharrtell? Would you be willing to share and commit it to this repo?

Otherwise, I can cobble something together. Would you be interested in a PR?

michaelkirk:
Fixes #194 

I used this to help verify correctness of https://github.com/flatgeobuf/flatgeobuf/pull/233


bjornharrtell:
Hmm.. strange, CircleCI complains "Could not find a usable config.yml because you deleted the CircleCI OAuth app.".

bjornharrtell:
Possibly a CircleCI incident - https://discuss.circleci.com/t/release-build-fails-because-deleted-the-circleci-oauth-app/45101

bjornharrtell:
Is the makefile worth it?

michaelkirk:
> Possibly a CircleCI incident - https://discuss.circleci.com/t/release-build-fails-because-deleted-the-circleci-oauth-app/45101

Ok let's try again! I've pushed up a no-op commit to trigger the build.

michaelkirk:
> Is the makefile worth it?

I can delete it if you'd like, but yes, I it's worthwhile, especially now that there are multiple targets.

If nothing else, I think it's a more likely place for people to find build steps.

In my own case, it would have saved me the time I spent digging around before I realized to check the `.circleci/config.yml`

bjornharrtell:
Good points. I'll merge now!

bjornharrtell (approved):

PR #238, opened 2022-10-24
Upgrade rust dependencies

bjornharrtell:


bjornharrtell:
Hi @pka please review if you have time. I would like to be more sure about the need to use the bytes accessor at https://github.com/flatgeobuf/flatgeobuf/blob/fedb59b9a0aa516a0c5cdffeb343dd4e23cbeb3d/src/rust/src/properties_reader.rs#L59.

pka (approved):
That's a major flatbuffers update with a single line comment in the ChangeLog...

The generated code changed from `fn properties(&self) -> Option<&'a [u8]>` to `fn properties(&self) -> Option<flatbuffers::Vector<'a, u8>>`, so the use of `bytes()` looks right to me. 

The other changes look fine, too and I trust the tests.

bjornharrtell:
They changed to date based versioning because they were unable to make sense of semver with all different lang bindings etc.

Issue #235, opened 2022-09-05
JS/TS: Remove logging from production code/builds


bjornharrtell:

bjornharrtell:
TypeScript tsc is missing capability to conditionally omit in compilation (https://github.com/microsoft/TypeScript/issues/3538).

A possible option is to switch to esbuild and use https://github.com/zziger/esbuild-ifdef

bjornharrtell:
esbuild is only suited for bundles not pure TypeScript transpilation... swc seems a better candidate but has no plugin for preprocess/ifdef. :(

PR #241, opened 2022-10-27
Use swc instead of tsc

bjornharrtell:
ref https://github.com/flatgeobuf/flatgeobuf/issues/235


bjornharrtell:
Won't do, it doesn't actually do type check.

PR #247, opened 2022-11-28
Optimize build write index performance

ChenyuWang98:
When the number of nodes is large, it takes longer and longer. The optimization effect is obvious. 
I think it can be merged into the master branch.

### DIFF
--- a/src/java/src/main/java/org/wololo/flatgeobuf/PackedRTree.java
+++ b/src/java/src/main/java/org/wololo/flatgeobuf/PackedRTree.java
@@ -6,11 +6,12 @@ import java.io.OutputStream;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.util.ArrayList;
-import java.util.Collections;
-import java.util.LinkedList;
+import java.util.Iterator;
 import java.util.List;
-import java.util.Objects;
 import java.util.Stack;
+import java.util.LinkedList;
+import java.util.Collections;
+import java.util.Objects;
 
 import org.locationtech.jts.geom.Envelope;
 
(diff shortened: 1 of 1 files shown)


bjornharrtell:
I would not expect an iterator to be faster than indexing?

ChenyuWang98:
image
The fact is that the effect is obvious

michaelkirk:
This is likely because the test is using a LinkedList: https://github.com/flatgeobuf/flatgeobuf/blob/master/src/java/src/test/java/org/wololo/flatgeobuf/PackedRTreeTest.java#L27

ChenyuWang98:
> 这可能是因为测试使用的是链表：[https ://github.com/flatgeobuf/flatgeobuf/blob/master/src/java/src/test/java/org/wololo/flatgeobuf/PackedRTreeTest.java#L27](https://github.com/flatgeobuf/flatgeobuf/blob/master/src/java/src/test/java/org/wololo/flatgeobuf/PackedRTreeTest.java#L27)


Yes, linked list is used

bjornharrtell:
Hmm I see. I can merge it but would like to avoid star imports `import java.util.*` can you please change that @ChenyuWang98?

ChenyuWang98:
> ` [db4b068](/flatgeobuf/flatgeobuf/pull/247/commits/db4b068a31cf169604afd0ddabe2fb919da46ca6)`


db4b068

ChenyuWang98:
> Hmm I see. I can merge it but would like to avoid star imports `import java.util.*` can you please change that @ChenyuWang98?


db4b068

bjornharrtell (approved):

bjornharrtell:
Thanks @ChenyuWang98

ChenyuWang98:
I think a java version 3.24.2 should be released

PR #272, opened 2023-07-07
Rust: Release 3.26.0 with updated dependencies

pka:


src/rust/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "flatgeobuf"
-version = "3.25.0"
+version = "0.9.0"
```
bjornharrtell, line 3: I made it 3.x some time ago and it's on crates.io so I don't think it makes sense to try and go back?

pka: Oh, I wans't aware of that. Is the idea to have the same version number in all languages? A difficulty for the Rust implementatiion could be different SemVer compatibility rules. According to https://doc.rust-lang.org/cargo/reference/semver.html each change in the public API is a major update, which means for crate version 3.25.0 (>1.0) going to 4.0.0, but for 0.8.0 (<1.0) it would have been 0.9.0.

bjornharrtell: It's more or less the same for npm, perhaps less officially but still. So, mabye I shot myself in the foot with attempting to use spec version av major version. But I guess I see nothing wrong with bumping major going forward if we need to make breaking changes with or without bumping other langs.

bjornharrtell: Food for thought is that flatbuffers was in a similar situation and decided to go with date scheme versioning even though that is not semver. Initially I thought it was a wired potentially bad decision but it has actually grown on me. I very much like semver but it's difficult to maintain when you have multi lang targets and multi layered compatibility (api vs wire protocol).

pka: We do also date scheme versioning in https://github.com/qwc-services, so I'm used to that.

bjornharrtell (approved):

Issue #274, opened 2023-07-18
Java PackedRTree inefficient streaming search


aaime:
The streaming search re-sorts the entire Stack it's using after each insertion, which makes it rewrite the entire backing data structure. A PriorityQueue is designed to insert new items in the right order without as much overhead.

bjornharrtell:
After some recollection and thinking I think that sort was added by mistake for some reason. I do not believe it is needed as the flatgeobuf index is guaranteed to be sorted already.

I've made https://github.com/flatgeobuf/flatgeobuf/pull/277 that removes the sort and replace Stack with Deque.

bjornharrtell:
I'll try to get this out in a patch release later today and then do downstream patches.

PR #275, opened 2023-07-18
#274 Java PackedRTree inefficient streaming search

aaime:
See ticket for details.

### DIFF
--- a/src/java/src/main/java/org/wololo/flatgeobuf/PackedRTree.java
+++ b/src/java/src/main/java/org/wololo/flatgeobuf/PackedRTree.java
@@ -9,10 +9,12 @@ import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 import java.util.NoSuchElementException;
+import java.util.PriorityQueue;
 import java.util.Stack;
 import java.util.LinkedList;
 import java.util.Collections;
 import java.util.Objects;
+import java.util.TreeSet;
 
 import org.locationtech.jts.geom.Envelope;
 
@@ -336,10 +338,10 @@ public class PackedRTree {
         double maxY = rect.getMaxY();
         ArrayList<Integer> levelEnds = generateLevelEnds(numItems, nodeSize);
         int numNodes = levelEnds.get(0);
-        Stack<StackItem> stack = new Stack<StackItem>();
-        stack.add(new StackItem(0, levelEnds.size() - 1));
-        while (stack.size() != 0) {
-            StackItem stackItem = stack.pop();
+        PriorityQueue<StackItem> queue = new PriorityQueue<>((a, b) -> (int) (a.nodeIndex - b.nodeIndex));
+        queue.add(new StackItem(0, levelEnds.size() - 1));
+        while (queue.size() != 0) {
+            StackItem stackItem = queue.poll();
             int nodeIndex = (int) stackItem.nodeIndex;
             int level = stackItem.level;
             boolean isLeafNode = nodeIndex >= numNodes - numItems;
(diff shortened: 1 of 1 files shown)


src/java/src/main/java/org/wololo/flatgeobuf/PackedRTree.java
```diff
@@ -336,10 +338,10 @@ public static SearchResult search(InputStream stream, int start, int numItems, i
         double maxY = rect.getMaxY();
         ArrayList<Integer> levelEnds = generateLevelEnds(numItems, nodeSize);
         int numNodes = levelEnds.get(0);
-        Stack<StackItem> stack = new Stack<StackItem>();
-        stack.add(new StackItem(0, levelEnds.size() - 1));
-        while (stack.size() != 0) {
-            StackItem stackItem = stack.pop();
+        PriorityQueue<StackItem> queue = new PriorityQueue<>((a, b) -> (int) (a.nodeIndex - b.nodeIndex));
+        queue.add(new StackItem(0, levelEnds.size() - 1));
```
michaelkirk, line 342: I don't disagree that a PriorityQueue should be faster than a stack in this case. But I'm curious now...

Since we're traversing the tree breadth first, that should be equivalent to just a regular queue - where we push onto the back and pop off the front. Right?

The [typescript implementation](https://github.com/flatgeobuf/flatgeobuf/blob/ec48353b169e41947a1923b9560494916ad5ed75/src/ts/packedrtree.ts#L144) actually use a straight-up queue like this.

Interestingly, the [CPP](https://github.com/flatgeobuf/flatgeobuf/blob/ec48353b169e41947a1923b9560494916ad5ed75/src/cpp/packedrtree.cpp#L341) and [rust](https://github.com/flatgeobuf/flatgeobuf/blob/ec48353b169e41947a1923b9560494916ad5ed75/src/rust/src/packed_r_tree.rs#L452) implementations use on ordered map - which actually seems wrong to me since [unordered_map doesn't guarantee any sorting](https://cplusplus.com/reference/unordered_map/unordered_map/begin/), right?

I might be missing something here though - I haven't looked into that closely.

aaime: I did not look into it closely either, but I can tell you that at the beginning I had the direction of comparison flipped and tests were not passing anymore (did not check if the test was too rigid or it was a legit failure).

The core of the change is switching from sorting Stack, which uses TimSort with a computational complexity of O(n log n), towards inserting in a priority queue, which should have a complexity of O(log n).

bjornharrtell: I wasn't aware Stack was sorted? Can't figure that from docs either (though it does say Deque should be used instead). The original implementation is from flatbush which most closely resembles what I implemented for TypeScript... that is, I don't think ordering is important here.

aaime: Stack by itself is not sorted. It was getting sorted after every insert, that was the bit that slowed everything down a lot.

src/java/src/main/java/org/wololo/flatgeobuf/PackedRTree.java
```diff
@@ -382,9 +384,8 @@ public static SearchResult search(InputStream stream, int start, int numItems, i
                 if (isLeafNode)
                     searchResult.hits.add(new SearchHit(indexOffset, pos - 1));
                 else
-                    stack.add(new StackItem(indexOffset, level - 1));
+                    queue.add(new StackItem(indexOffset, level - 1));
             }
-            stack.sort((StackItem a, StackItem b) -> (int) (b.nodeIndex - a.nodeIndex));
```
aaime, line 387: This is the killer line. That said, Stack is one of the JDK 1.0 data structures. It's vastly outdated. If you just need something to beheave like a stack ,with no ordering, pick a Deque implementation instead

bjornharrtell: I now think this sort was a mistake, the items are already guaranteed to be in order.

aaime:
Superceded by #277

PR #277, opened 2023-07-19
Replace Stack with Deque and remove redundant sort

bjornharrtell:
ref #274


bjornharrtell:
@aaime that is indeed curious.

bjornharrtell:
Ah.. the pop operation is expensive for ArrayDeque. I will change to LinkedList.

aaime:
I see.. but also my fault. The tests I did earlier contained other improvements in the Geotools rendering subsystem that I did not include in this last test run.

Run it again with those improvements as well, and with the linked list: 32 seconds. Noice!

bjornharrtell:
Sounds good enough for me to merge. :)

PR #278, opened 2023-07-19
rust: fix inconsistent result ordering

michaelkirk:
In the `search` method, we're using a HashMap as a "queue" but HashMaps don't
maintain ordering. So we sometimes see results returned in different order.

Since we're traversing the tree breadth first, we don't need a sorting data
structure - we can just add things to the back of a proper queue and pull from
the front. That is enough to ensure we're always getting the lowest index.

In the `streamSearch` method we were using a BinaryHeap which sorts based on index.
This behaved correctly, in that results were always in the correct order, but
it is unnecessarily slow/complicated.

In the HTTPSearch we were already using a VecDeque, so no changes there.

Related discussion occurred in #277, #274

### DIFF
--- a/src/rust/src/packed_r_tree.rs
+++ b/src/rust/src/packed_r_tree.rs
@@ -7,8 +7,7 @@ use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
 use geozero::error::{GeozeroError, Result};
 #[cfg(feature = "http")]
 use http_range_client::BufferedHttpRangeClient;
-use std::cmp::Reverse;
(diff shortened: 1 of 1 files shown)


bjornharrtell:
Seems sensible. Not sure why CircleCI is unhappy.

michaelkirk:
I originally pushed up a pretty stale branch. 

I rebased and force pushed, assuming it would magically retrigger the build, but I'm not convinced that it did. Just to maybe make things easy, I'm going to close this PR and open a new one to see if it makes circle CI happy.

PR #279, opened 2023-07-19
rust: fix inconsistent result ordering

michaelkirk:
(supersedes #278)

In the `search` method, we're using a HashMap as a "queue" but HashMaps don't
maintain ordering. So we sometimes see results returned in different order.

Since we're traversing the tree breadth first, we don't need a sorting data
structure - we can just add things to the back of a proper queue and pull from
the front. That is enough to ensure we're always getting the lowest index.

In the `streamSearch` method we were using a BinaryHeap which sorts based on index.
This behaved correctly, in that results were always in the correct order, but
it is unnecessarily slow/complicated.

In the HTTPSearch we were already using a VecDeque, so no changes there.

Related discussion occurred in #277, #274

### DIFF
--- a/src/rust/src/packed_r_tree.rs
+++ b/src/rust/src/packed_r_tree.rs
@@ -7,8 +7,7 @@ use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
 use geozero::error::{GeozeroError, Result};
 #[cfg(feature = "http")]
 use http_range_client::BufferedHttpRangeClient;
-use std::cmp::Reverse;
-use std::collections::{BinaryHeap, VecDeque};
+use std::collections::VecDeque;
 use std::io::{Cursor, Read, Seek, SeekFrom, Write};
 use std::mem::size_of;
 use std::{cmp, f64, u64, usize};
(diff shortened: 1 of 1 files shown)


bjornharrtell (approved):

michaelkirk:
>> Seems sensible. Not sure why CircleCI is unhappy.
>
> Just to maybe make things easy, I'm going to close this PR and open a new one to see if it makes circle CI happy.

Drat. CI is still angry.

michaelkirk:
Do you wan to try this @bjornharrtell?

https://support.circleci.com/hc/en-us/articles/12504745606171-Why-am-I-seeing-the-Could-not-find-a-usable-config-yml-you-may-have-revoked-the-CircleCI-OAuth-app-message


(Admittedly it's strange that it was working so recently... but I guess that's what happens when things break...)

bjornharrtell:
But is it not more likely your connection to CircleCI since it's your branch?

bjornharrtell:
"CircleCI will use the author's credentials in order to fetch the config.yml file"

bjornharrtell:
I've invited you to the org.

michaelkirk:
Oh perhaps! I didn't realize I needed to make an account on some other service for CI to run. 

Screenshot 2023-07-19 at 11 46 13

Sorry to be a pain, but I can't give circleci read/write access to all my repositories just to get this CI green.

If you'd like to copy these changes into a branch in this repository and open the PR yourself maybe things would run?

bjornharrtell:
Understandably. :) I'll just merge, this looks fine to me.

PR #288, opened 2023-08-21
Add test for single ring polygon

pka:
I tried to reproduce the bug fixed in #287. But I couldn't find a polygon in the test data encoded with `ends` == null.

https://github.com/flatgeobuf/flatgeobuf/blob/595a19f7237d9ebc899f4021ca1ebb84fa1b760f/src/fbs/feature.fbs#L6

@bjornharrtell do you have one?

### DIFF
--- a/src/rust/tests/read.rs
+++ b/src/rust/tests/read.rs
@@ -377,6 +377,27 @@ fn point_layer() -> Result<()> {
 }
 
 #[test]
+/// Test single ring polygons encoded with `ends` == null
+fn single_ring() -> Result<()> {
+    let mut filein = BufReader::new(File::open("../../test/data/poly00.fgb")?);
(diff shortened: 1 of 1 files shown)


bjornharrtell:
@pka hmm not readily. But makes me wonder how it's implemented in GDAL, mabye it always sets it.

pka:
I was wrong. The code is going into the `geometry.ends().is_none()` branch but I couln't find a wrong output yet. Investigating further...

pka:
The bug was not resulting in wrong geometries, it was rellay "just" passing a wrong size to `linestring_begin` which ends up in reserving to much memory in most writers. It did result in invalid WKB output and maybe wrong GEOS output though.

bjornharrtell:
@pka this this is good to merge?

pka:
> @pka this this is good to merge?

Yes

Issue #94, opened 2020-12-22
rust http client idea: stream rather than chunk 


michaelkirk:
The current streaming strategy is based on chunks: 

1. request a chunk
2. iterate over the chunk
3. request a chunk
4. iterate over the chunk
5. ...

It seems like we could increase throughput, and in particular reduce the latency-to-first-feature, by utilizing a continuous byte stream.

Our http lib, reqwest supports [`response.byte_stream`](https://docs.rs/reqwest/0.10.10/reqwest/struct.Response.html#method.bytes_stream) if you enable the `stream` feature which might be useful.

I'm personally not familiar with creating Futures and consuming Stream apis in rust, but this seems conceptually straight forward for the feature-fetching portion of the http reader.

There *might* be some way to also utilize it for the header+index downloads as well, but that seems a more complicated cost to benefit ratio.

michaelkirk:
If anyone has any thoughts on this, I'm having a go at it.

bjornharrtell:
While I'm still (forever?) a rust-noob it sounds like a good idea to me. :) @pka?

pka:
My gut feeling says that only the first point, reduce the latency-to-first-feature, will be a measurable win. And then we need also a streaming FGB client, which does e.g. render the first feature immediately, to make the performance win also measurable for the end user.
As an alternative, experimenting with https://sozip.org/ could also bring a performance improvement for http access and would solve one downside of FGB, missing compression.

michaelkirk:
I would expect measurable wins anywhere the cost of a waiting on a request roundtrip is measurably faster than transmitting <chunk size> (currently 1MB). 

Conversely, I'd expect no real impact on small files or lower bandwidth connections, especially if they have relatively low latency. 

Some motivating (but naive) math:

<img width="1259" alt="Screenshot 2023-09-01 at 12 00 19" src="https://github.com/flatgeobuf/flatgeobuf/assets/217057/d1980975-f033-4d1a-8026-af402de5cec4">

check my math: https://docs.google.com/spreadsheets/d/1eu35b1mFwKTN-LzpmJFjwC4n4uP9AwfuPxuSh5OwO2k/edit#gid=0

This is all just theoretical numbers, I'm sure it won't be that good, but I'm reasonably optimistic that  there will be more wins than only "time to first feature", especially for larger files on fast connections, without requiring users to hand-tune their chunk size. It also has the upside that you don't have to hold chunk_size in memory all at once.

[...]

PR #293, opened 2023-09-06
[rust] benchmarks for reading via HTTP

michaelkirk:
I've been digging into the http reader's performance characteristics as part of https://github.com/flatgeobuf/flatgeobuf/issues/94

For (somewhat) reproducible results, a local http server is used to host the files.
For (somewhat) realistic measurements, the server induces artificial latency and bandwidth constraints.

There's still some noise in the measurements, but it's much better to the point of being actually useful in the work I've been doing. For example, I still regularly see 1-5% differences in measurements. But without the local server, I was almost always seeing swings of 80% or more.

### DIFF
--- a/src/rust/Cargo.toml
+++ b/src/rust/Cargo.toml
@@ -30,15 +30,21 @@ geozero = { version = "0.11.0", default-features = true }
 seek_bufread = "1.2.2"
 rand = "0.8.5"
 hex = "0.4.3"
-criterion = "0.5.1"
-tokio = { version = "1.30.0", default-features = false, features = ["macros"] }
+criterion = { version = "0.5.1", features = ["async_tokio"] }
+tokio = { version = "1.30.0", features = ["rt-multi-thread"] }
 # One test needs SSL support; just use the default system bindings for that.
 reqwest = { version = "0.11.18", default-features = true }
 geo-types = "0.7.11"
+yocalhost = "0.1.0"
 
 [[bench]]
 name = "read"
 harness = false
+
+[[bench]]
+name = "http_read"
+harness = false
+
 [[bench]]
 name = "geojson"
 harness = false
--- a/src/rust/benches/http_read.rs
+++ b/src/rust/benches/http_read.rs
@@ -1,0 +1,74 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+
+use flatgeobuf::HttpFgbReader;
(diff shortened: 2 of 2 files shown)


src/rust/Cargo.toml
```diff
@@ -30,15 +30,21 @@ geozero = { version = "0.11.0", default-features = true }
...
 rand = "0.8.5"
 hex = "0.4.3"
-criterion = "0.5.1"
-tokio = { version = "1.30.0", default-features = false, features = ["macros"] }
+criterion = { version = "0.5.1", features = ["async_tokio"] }
+tokio = { version = "1.30.0", features = ["rt-multi-thread"] }
 # One test needs SSL support; just use the default system bindings for that.
 reqwest = { version = "0.11.18", default-features = true }
 geo-types = "0.7.11"
+yocalhost = "0.1.0"
```
michaelkirk, line 38: full disclosure, this is something I made: https://github.com/michaelkirk/yocalhost

If there's an existing better solution I'd be happy to use it. I found some none-rust solutions, but the couple I tried didn't immediately work out or involved GUI setup.

src/rust/benches/http_read.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+fn criterion_benchmark(c: &mut Criterion) {
+    use std::time::Duration;
+    use yocalhost::ThrottledServer;
+
+    let port = 8001;
+
+    // Apply limits to simulate an "average" connection for benchmarks
+    let latency = Duration::from_millis(100);
```
michaelkirk, line 44: I'm happy to bikeshed these values. I tried to pick something average.

pka: Schould probably be an env variable.

bjornharrtell:
While I can't chip in with much here I very much like the initiative. :)

src/rust/benches/http_read.rs
michaelkirk: Good idea.

Now you can do something like this:
```
FGB_BENCH_BYTES_PER_SEC=125000000 FGB_BENCH_LATENCY_MS=50 cargo bench --bench http_read
```

michaelkirk:
Is this something you're interested in merging @pka? LMK if you're waiting for further changes from me.

pka (approved):
:+1:

PR #297, opened 2023-09-08
[rust] update http-range-client to ensure tls support

michaelkirk:
The previous version of http-range-client didn't enable tls support.

Because features in rust are additive, it's possible that downstream users had enabled tls via some other dependency, but we should make sure it's always supported.

### DIFF
--- a/src/rust/Cargo.toml
+++ b/src/rust/Cargo.toml
@@ -19,7 +19,7 @@ http = ["http-range-client", "bytes"]
 flatbuffers = "23.5.26"
 byteorder = "1.4.3"
 geozero = { version = "0.11.0", default-features = false }
-http-range-client = { version = "0.6.0", optional = true }
+http-range-client = { version = "0.7.0", optional = true }
 bytes = { version = "1.4.0", optional = true }
 log = "0.4.19"
 fallible-streaming-iterator = "0.1.9"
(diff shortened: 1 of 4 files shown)


src/rust/src/http_reader.rs
```diff
@@ -65,29 +65,22 @@ impl HttpFgbReader<Initial> {
         // fetch an extra kb rather than have to issue a second request.
         let assumed_header_size = 2024;
         let min_req_size = assumed_header_size + prefetch_index_bytes;
+        client.set_min_req_size(min_req_size);
         debug!("fetching header. min_req_size: {min_req_size} (assumed_header_size: {assumed_header_size}, prefetched_index_bytes: {prefetch_index_bytes})");
 
-        let bytes = client
-            .get_range(0, 8, min_req_size)
```
michaelkirk, line 71: Note the min_req_size API changed. 

(I'm sure you're aware @pka 😄)

bjornharrtell (approved):

src/rust/Cargo.toml
```diff
@@ -19,7 +19,7 @@ http = ["http-range-client", "bytes"]
 flatbuffers = "23.5.26"
 byteorder = "1.4.3"
 geozero = { version = "0.11.0", default-features = false }
-http-range-client = { version = "0.6.0", optional = true }
```
pka, line 22: Use 0.7.1, since 0.7.0 had a major bug.

PR #301, opened 2023-09-11
Avoid waiting one second before requesting data

tschaub:
It looks like the OpenLayers filtering examples currently use `_.debounce` when requesting updated data.  The Leaflet and Maplibre examples use `_.throttle` in this same case.  The result is that the OpenLayers example waits 1 second before fetching data while the other examples fetch it immediately (and then avoid fetching more than once per second).

This change updates the OpenLayers example to use `_.throttle` instead of `_.debounce` to match the behavior of the other examples.

In addition, I've set `useSpatialIndex: false` on the vector source for the examples that already filter the data by extent.  In this case, the spatial index is not needed on the source.

Here is the behavior after this change (no delay before fetching data):

after

And here is the behavior before this change (one second delay before fetching data):

before

### DIFF
--- a/examples/openlayers/filtered.html
+++ b/examples/openlayers/filtered.html
@@ -77,7 +77,8 @@
                 };
             }
 
-            const results = new ol.source.Vector();
+            // no need for the spatial index since this is only rendering a small extent
+            const results = new ol.source.Vector({useSpatialIndex: false});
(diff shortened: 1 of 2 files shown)


bjornharrtell:
@tschaub good catch! And thanks, will merge.

bjornharrtell (approved):

PR #304, opened 2023-09-14
[rust] refactor: Extract a FeatureIter struct rather than track State param

michaelkirk:
The State parameter was introduced as a type check to make sure we weren't calling methods at the wrong time - e.g. we can't iterate over a reader before we've selected features.

Instead, we now extract all the state related to iteration into the FeatureIter, achieving the same type checking in a more conventional way (and with a little less code).

Also, the previous mechanism worked only for method calls, but not for field access.  Some of the structs fields were only safe to access when the state had a certain value - like feature_count only gets populated *after* selection. This was already protected by the public method, but now it's also not possible to incorrectly access it via the field.

This might be controversial, since there's no immediate behavioral benefit.

I personally think it stands alone as a small understandability improvement, but TBH my real motivation for this PR is to minimize the diff for some [efficiency experiments](https://github.com/michaelkirk/flatgeobuf/tree/mkirk/streaming-reader) I'd like to start upstreaming. Since the new behavior pertains mostly to the selection (select_all, select_bbox), most of the new behavior ends up living on this newly extracted FeatureIter rather than on the Reader.

### DIFF
--- a/src/rust/src/file_reader.rs
+++ b/src/rust/src/file_reader.rs
@@ -2,7 +2,6 @@ use crate::feature_generated::*;
 use crate::header_generated::*;
 use crate::packed_r_tree::{self, PackedRTree};
 use crate::properties_reader::FgbFeature;
-use crate::reader_state::*;
 use crate::{check_magic_bytes, HEADER_MAX_BUFFER_SIZE};
 use fallible_streaming_iterator::FallibleStreamingIterator;
 use geozero::error::{GeozeroError, Result};
@@ -11,7 +10,16 @@ use std::io::{self, Read, Seek, SeekFrom};
 use std::marker::PhantomData;
 
 /// FlatGeobuf dataset reader
-pub struct FgbReader<R, State = Initial> {
+pub struct FgbReader<R> {
+    reader: R,
+    /// FlatBuffers verification
+    verify: bool,
+    // feature reading requires header access, therefore
+    // header_buf is included in the FgbFeature struct.
+    fbs: FgbFeature,
+}
+
+pub struct FeatureIter<R, S> {
     reader: R,
     /// FlatBuffers verification
     verify: bool,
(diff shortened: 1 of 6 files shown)


pka:
Looks great, after skimming quickly through! Will do a proper review tomorrow.

src/rust/src/file_reader.rs
```diff
@@ -28,13 +36,20 @@ pub struct FgbReader<R, State = Initial> {
     cur_pos: u64,
     /// All features read or end of file reached
     finished: bool,
-    /// Reader state
-    state: PhantomData<State>,
+    /// Whether or not the underlying reader is Seek
+    seekable_marker: PhantomData<S>,
+}
```
pka, line 41: Do you even need the `read_trait` parameter?

Isn't `R: Read + Seek` like in
```
impl<R: Read + Seek> FgbReader<R> {
    pub fn select_all(mut self) -> Result<FeatureIter<R>> {
```
already enough?

pka (changes requested):
This is much more elegant than the previous implementation!

Please check my comment, but I'm happy to approve/merge also as it is.

src/rust/src/file_reader.rs
michaelkirk: I'd be happy to get rid of it, but I don't think we can.

Ultimately we want to have two different  `impl FallibleStreamingIterator` declarations for the FgbReader — one for `Read` and another one optimized for `Read + Seek`.

Because `Read + Seek` is a subtype of `Read`, we have to do something to make sure the implementations aren't overlapping, otherwise you'll see something like:

```
error[E0119]: conflicting implementations of trait `MyTrait` for type `MyStruct<_>`
   --> src/file_reader.rs:255:1
    |
249 | impl<R: Read> MyTrait for MyStruct<R> {
    | ------------------------------------- first implementation here
...
255 | impl<R: Read + Seek> MyTrait for MyStruct<R> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `MyStruct<_>`
```

Because we want a distinct `impl FallibleStreamingIterator` for the thing returned by `select_all` vs `select_all_seq`, they must have distinct types.

We could introduce some new struct, e.g. `FeatureIterSeekable` and `FeatureIterNotSeekable`, but instead we return `FeatureIter<_, Seekable>`| `FeatureIter<_, NotSeekable>`.

I think this approach makes sharing common functionality a little easier. There's still a little duplicated code with this approach, but I think it's as minimal as it could be. I actually think it's kind of clever, and I don't think I would have come up with the idea if you hadn't introduced a similar concept in 30af03d0811607d7b40166e9bbd5efb69c830557, so thanks for the inspiration. =)

pka (approved):

PR #306, opened 2023-09-14
[go] add CI

michaelkirk:


michaelkirk:
I don't use go much. @bruno-ga does this seem right? Anything else we should be running?

I did verify that CI fails if I break a test - e.g.  https://github.com/michaelkirk/flatgeobuf/actions/runs/6190262736/job/16806141412

bjornharrtell (approved):
Thanks @michaelkirk!

bruno-ga:
> I don't use go much. @bruno-ga does this seem right? Anything else we should be running?
> 
> I did verify that CI fails if I break a test - e.g. https://github.com/michaelkirk/flatgeobuf/actions/runs/6190262736/job/16806141412

Yep, this looks reasonable. The test coverage is not great but that is not your fault. I will continue working on this. Next steps are adding more tests and a FeatureGenerator example (maybe a geo-json one to match other languages). FWIIW, at work we use shapefiles.

PR #308, opened 2023-09-18
[go] Only allocate memory when needed

bruno-ga:
- Even when not copying the data for the index (only relevant for memory
   mapped flatgeobuf files, we were allocating the memory to hold the
   copy.


bjornharrtell (approved):

bruno-ga:
Sorry, I had to force push to clean up my remote branch. Can you take another look? The actual change is like 2 lines or so. :)

Issue #283, opened 2023-08-10
Resolve circular dependency of Rust implemention with geozero


pka:
We have a long standing maintenance problem because of the circular dependency between Rust Flatgeobuf and geozero. I've discussed that with @nyurik and we see the following solutions:

1. Move rust/flatgeobuf to the geozero repo
2. Extract the geozero dependent part of rust/flatgeobuf into a geozero driver
3. Remove all flatgeobuf dependencies from the geozero repo

Pros/Cons:
* Option 1 is technically the easiest. We just had to give @bjornharrtell all the required Github permissions.
* Option 2 gets at least difficult in the test dependencies. To make readable tests we need to convert from/to FlatGeoBuf and the only option for that is using geozero. We could move them also to geozero, but that's not nice.
* Option 3 is also quite a lot of work, because flatgeobuf is used in many places. We have e.g. geozero-cli which converts other geozero supported formats from/to Flatgeobuf, we have benchmarks and so on.

So before going into more detail, is there an option you would rule out, @bjornharrtell?

nyurik:
Thanks @pka for an awesome write-up!  I agree that option 1 is by far the simplest and safest from the code quality and maintenance perspective.

Fundamentally, we have a bad situation of `flatgeobuf` having tons of dependencies on `geozero`, which in turn depends on `flatgeobuf` in many unit tests, benchmark, and `geozero-cli` code.  This means that any bump in minor or major version causes CI local builds to fail, especially when both the geozero AND test/cli/bench code is changed at the same time.

Geozero relies on 3rd party libraries to handle the low level processing of the geo data. For example MVT relies on protobuf, and GeoJSON on serde.  Yet, `flatgeobuf` usage is unusual -- instead of geozero relying on flatgeobuf, it is the other way around.  In theory, `flatgeobuf` should offer a low level binary stream encoding/decoding, i.e. being similar to serde or protobuf, and GeoZero should do the higher level processing, e.g. iterate over the stream of data and convert it to the common internal feature representation for geometries and properties.

This essentially is the option 2, and I think it is a better path long term. Yet, to get there, most of the refactoring would need to be done in close sync inside a single repo with a common CI. Thus, I would recommend starting with option 1, and possibly try to move to option 2.

P.S note that `fgbutil` does not need to move, only the `flatgeobuf` crate and probably bench/fuzz stuff.

bjornharrtell:
[...]

PR #310, opened 2023-09-27
Introduce local Error type rather than using Geozero::GeozeroError

michaelkirk:
## Motivation

Currently we have a circular dependency between the geozero and flatgeobuf crates. Some related discussion in
https://github.com/flatgeobuf/flatgeobuf/issues/283

This starts to detangle us from geozero by introducing our own error type and using that for non-geozero code. It also allows us to have an error enum more suited for our use cases.

## What changed

1. Introduce a new `flatgeobuf::Error` type and use it for code that doesn't depend on Geozero.
2. Relocate code that depends on Geozero into separate `geozero_integration` modules. This code still returns GeozeroError.
3. For tests and benchs, which might return either `flatgeobuf::Error` or `geozero::GeozeroError` and mapping between them is verbose, use `Box<dyn std::error::Error>`. Alternatively, I'd also be OK with using `unwrap` everywhere in test code, but I was trying not to rock the boat too much.

### DIFF
--- a/src/rust/src/error.rs
+++ b/src/rust/src/error.rs
@@ -1,0 +1,45 @@
+use flatbuffers::InvalidFlatbuffer;
+use std::fmt::{Display, Formatter};
+
+#[derive(Debug)]
+pub enum Error {
+    Malformed(&'static str),
+    IO(std::io::Error),
+    InvalidFlatbuffer(InvalidFlatbuffer),
+    #[cfg(feature = "http")]
+    HttpClient(http_range_client::HttpError),
+}
+pub type Result<T> = std::result::Result<T, Error>;
+
+impl Display for Error {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Error::Malformed(description) => description.fmt(f),
+            Error::IO(io) => io.fmt(f),
+            Error::InvalidFlatbuffer(invalid_flatbuffer) => invalid_flatbuffer.fmt(f),
+            #[cfg(feature = "http")]
+            Error::HttpClient(http_client) => http_client.fmt(f),
+        }
+    }
+}
+
+impl std::error::Error for Error {}
+
+impl From<std::io::Error> for Error {
+    fn from(value: std::io::Error) -> Self {
+        Self::IO(value)
+    }
+}
+
+impl From<InvalidFlatbuffer> for Error {
+    fn from(value: InvalidFlatbuffer) -> Self {
+        Error::InvalidFlatbuffer(value)
+    }
+}
+
+#[cfg(feature = "http")]
+impl From<http_range_client::HttpError> for Error {
+    fn from(value: http_range_client::HttpError) -> Self {
+        Error::HttpClient(value)
+    }
+}
--- a/src/rust/src/error.rs
+++ b/src/rust/src/error.rs
@@ -3,7 +3,7 @@ use std::fmt::{Display, Formatter};
 
 #[derive(Debug)]
 pub enum Error {
-    MissingMagicNumber,
+    MissingMagicBytes,
     NoIndex,
     #[cfg(feature = "http")]
     HttpClient(http_range_client::HttpError),
@@ -16,7 +16,7 @@ pub type Result<T> = std::result::Result<T, Error>;
 impl Display for Error {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         match self {
-            Error::MissingMagicNumber => "Missing magic bytes. Is this an fgb file?".fmt(f),
+            Error::MissingMagicBytes => "Missing magic bytes. Is this an fgb file?".fmt(f),
             Error::NoIndex => "Index missing".fmt(f),
             #[cfg(feature = "http")]
             Error::HttpClient(http_client) => http_client.fmt(f),
--- a/src/rust/src/error.rs
+++ b/src/rust/src/error.rs
@@ -3,22 +3,26 @@ use std::fmt::{Display, Formatter};
 
 #[derive(Debug)]
 pub enum Error {
-    Malformed(&'static str),
-    IO(std::io::Error),
-    InvalidFlatbuffer(InvalidFlatbuffer),
+    MissingMagicNumber,
+    NoIndex,
     #[cfg(feature = "http")]
     HttpClient(http_range_client::HttpError),
+    IllegalHeaderSize(usize),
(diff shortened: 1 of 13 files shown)


michaelkirk:
The corresponding geozero PR is here: https://github.com/georust/geozero/pull/175

pka:
I quickly skimmed through the changes, and have general comments before going into detail
1. Error type: :+1: :+1: 
2. I'm not sure about that. I recently grepped for flatgeobuf in the geozero "core" crate and found so few hits, that I think removing flatgeobuf from the geozero crate is the way to #283. I'll add a comment there with more details.
3. No comments yet

michaelkirk:
Rebased, fixed an old lint, and realized the FallibleStreamingIterator need not return a GeozeroError.

> I'm not sure about that. I recently grepped for flatgeobuf in the geozero "core" crate and found so few hits, that I think removing flatgeobuf from the geozero crate is the way to https://github.com/flatgeobuf/flatgeobuf/issues/283. I'll add a comment there with more details.

I'll respond in that thread in hopes of centralizing the discourse. I'm hopeful that at least some of this PR is useful regardless of which direction we take on the "big picture" of resolving the circular dependency.

michaelkirk:
@pka - would you prefer that I revert the `mod geozero_integration` for now?

src/rust/src/error.rs
```diff
@@ -0,0 +1,45 @@
+use flatbuffers::InvalidFlatbuffer;
+use std::fmt::{Display, Formatter};
+
+#[derive(Debug)]
+pub enum Error {
+    Malformed(&'static str),
```
pka, line 6: Are there many different error texts? @nyurik argued that we shouldn't use freetext messages for errors at all.

src/rust/src/file_reader.rs
```diff
@@ -351,6 +301,117 @@ impl<R: Read + Seek> FallibleStreamingIterator for FeatureIter<R, Seekable> {
     }
 }
 
+mod geozero_integration {
```
pka, line 304: I would prefer `geozero_api`

src/rust/src/file_writer.rs
```diff
@@ -365,129 +319,186 @@ impl<'a> FgbWriter<'a> {
     }
 }
 
-impl FeatureProcessor for FgbWriter<'_> {
-    fn feature_end(&mut self, _idx: u64) -> Result<()> {
-        self.write_feature()
+mod geozero_integration {
```
pka, line 322: `geozero_api`

src/rust/src/http_reader.rs
```diff
@@ -199,39 +185,40 @@ impl AsyncFeatureIter {
...
-    /// Read and process all selected features
-    pub async fn process_features<W: FeatureProcessor>(&mut self, out: &mut W) -> Result<()> {
-        out.dataset_begin(self.fbs.header().name())?;
-        let mut cnt = 0;
-        while let Some(feature) = self.next().await? {
-            feature.process(out, cnt)?;
-            cnt += 1;
+}
+
+mod geozero_integration {
```
pka, line 204: same

src/rust/src/packed_r_tree.rs
```diff
@@ -886,6 +893,7 @@ fn tree_processing() -> Result<()> {
     }
     let tree = PackedRTree::build(&nodes, &extent, PackedRTree::DEFAULT_NODE_SIZE)?;
     let mut fout = BufWriter::new(tempfile()?);
-    tree.process_index(&mut GeoJsonWriter::new(&mut fout))?;
+    tree.process_index(&mut GeoJsonWriter::new(&mut fout))
+        .unwrap();
```
pka, line 897: unwrap?

src/rust/tests/read.rs
```diff
@@ -328,7 +317,7 @@ fn reader_type() -> Result<()> {
     fn get_feature_reader<R: Read + Seek>(
         reader: R,
     ) -> Result<FeatureIter<R, reader_trait::Seekable>> {
-        FgbReader::open(reader)?.select_all()
+        Ok(FgbReader::open(reader)?.select_all()?)
```
pka, line 320: This is the same as before, right? Is that better style?

pka (changes requested):
Apart from a forgotten `unwrap`, everything is a an improvement over the current state, so fine fore merging. A few nitpicks we can discuss if you like.

src/rust/src/error.rs
nyurik: @pka I think `static str` is better than `String` (no allocation, much faster errors, esp when no error logging is needed), but I also think that a dedicated error type is better than `static str` because the user can make decisions based on the error type, and the compiler will ensure the stability of those errors (as oppose to doing brittle error string comparison).

src/rust/tests/read.rs
nyurik: `?` here might call an `Into` trait, whereas simply returning a `Result` might not work in case the error type is different

src/rust/src/error.rs
michaelkirk: I think freetext messages can be OK if you don't intend for the user to handle the specific texts differently.

That said, I think in this case Malformed was doing more work than it should. See b7e7f73

Are the unwrap's controversial? We're already asserting the same things that guarantee the invariant within the `generate_bounds` method so I don't think it introduces any crash in practice. I could leverage the typesystem to return some kind of "NonEmptyVec", but I'm not sure it's worth it.

nyurik:
I'm a bit confused - why are you using boxed errors?  In my understanding, this slower approach is mostly used in apps (i.e. `anyhow` crate does that internally), whereas most libs rely on `thiserror` style - where each error is statically typed

michaelkirk:
Firstly, note that the boxed errors are only used in test code, so I'm not overly concerned about performance there. 

This test code represents a place where the two dependencies (geozero vs fgb) haven't been detangled - where the code could be propagating either GeozeroError or flatgeobuf::Error.

nyurik:
ah, ok, yes, test code is fine i guess (would it perhaps make sense to simply rely on anyhow then to re-use existing pattern?  Definitely not a blocker though).  Also another possible solution might be to use this (untested):

```rust
#[derive(thiserror)]
enum MixedError {
  #[error("GeoZero error {0}"),
  GzError(#[from] GeozeroError),
  #[error("FlatGeoBuf error {0}"),
  FgbError(#[from] flatgeobuf::Error),
}
```

pka (approved):

PR #315, opened 2023-10-01
[go] Fix invalid array length

shihanng:
We are getting the following errors during compilation when targetting a 32-bit system (e.g., tinygo build -target=wasi ...)

```
../src/go/index/packed_r_tree.go:144:15: invalid array length 1 << 31 (untyped int constant 2147483648)
../src/go/index/packed_r_tree.go:237:19: invalid array length 1 << 31 (untyped int constant 2147483648)
```

This commit fixes the issues above.

This [SO answer](https://stackoverflow.com/a/28552459) is the only reference I can find regarding the maximum array size.

> The size 1<<31 - 1 is the largest possible size any slice in Go can have.

I also confirm that with this change, I can compile successfully.

### DIFF
--- a/src/go/index/packed_r_tree.go
+++ b/src/go/index/packed_r_tree.go
@@ -137,7 +137,7 @@ func (r *PackedRTree) Size() uint64 {
 // with NewPackedRTreeFromData.
 func (r *PackedRTree) Write(w io.Writer) (int, error) {
 	// Fast conversion from  []NodeItems to []byte.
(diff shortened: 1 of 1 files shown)


bjornharrtell:
Thanks, seems like a no brainer fix for me so I'll merge.

bjornharrtell (approved):

bruno-ga:
FWIIW, yes, the change should be ok. That is what you get when you only work with 64 bit platforms. :) Thanks for fixing it.

PR #316, opened 2023-10-01
JS/TS: Use DataView and getBigUint64 instead of high/low bits read

bjornharrtell:


michaelkirk (approved):
LGTM!

Were you only trying to simplify this code a little bit or do you have an end goal of exposing more of the library to BigInt so we can support larger than 52 bit integers?

bjornharrtell:
Simplification only 🙂

PR #318, opened 2023-10-01
[rust] Skip redundant sort - list is sorted by construction.

michaelkirk:


bjornharrtell (approved):
I think this is valid. I don't think we have the same "issue" in the ts implementation, but let me know if you think otherwise.

PR #319, opened 2023-10-02
[rust] Batch nearby HTTP requests for bbox features

michaelkirk:
~~DRAFT: because it's based on #318 . Please review that first.~~ Merged!

When reading FGB over HTTP, most time is spent in the network, as opposed to local processing.

To minimize network round trips, we currently "over fetch" whenever we're getting feature data, in hopes of obviating a future request. For "select_all" queries this works great, because eventually we're going to need all the data anyway.

For bbox queries, we specifically don't want *all* the data.  Thanks to hilbert ordering, spatially nearby features are typically nearby on disk so we do OK with the same strategy, but inevitably we end up fetching data we don't end up using. 

The hilbert index contains the start location of every feature. We can infer the length of each feature by knowing where the _next_ feature starts. So we actually have all the information to fetch precisely the feature data we need for a given bbox.

So we could precisely fetch every feature individually to minimize bandwidth, but because network round trips are expensive, it still makes sense to "waste" some bytes and combine these feature requests. But now we can quantify exactly how many bytes we're willing to waste to obviate a separate request, and we don't bother fetching extra data when we know it's not useful.

I made a similar PR for JS/TS a while back (though that one also does concurrent fetching): https://github.com/flatgeobuf/flatgeobuf/pull/135

## Time improvements

Recall that our benchmarks are run against a simulated network which has 100ms latency and 50Mbps bandwidth.

```
medium select_bbox
time:   [636.61 ms 637.89 ms 639.27 ms]
change: [-41.733% -41.515% -41.296%] (p = 0.00 < 0.05)
Performance has improved.

medium bbox -> geojson
time:   [661.92 ms 663.29 ms 664.79 ms]
change: [-40.336% -40.172% -39.995%] (p = 0.00 < 0.05)
Performance has improved.
```

The rest of the benches are unaffected. Since `select_all`'s fetch strategy didn't change and bbox requests on small files already fit into the over-fetch of our initial header request, this is exactly as expected.

## Network improvements

In the above example, byte consumption was reduced from 4.25MB to 1.78MB

Before feature batching:
```
req_count: 6, body_bytes_sent: 4253968
```

With feature batching:
```
req_count: 6, body_bytes_sent: 1775808
```

Note that, in this particular case from our existing benches, the total number of requests were not reduced. But if the feature batches are close enough, and we're within our maximum request size, we will sometimes consolidate requests.

Note we still want to have _some_ maximum request size, because our current chunking approach stores the entire response in memory.

### DIFF
--- a/src/rust/src/http_reader.rs
+++ b/src/rust/src/http_reader.rs
@@ -1,12 +1,17 @@
 use crate::feature_generated::*;
 use crate::header_generated::*;
-use crate::packed_r_tree::{self, NodeItem, PackedRTree};
+use crate::packed_r_tree::{HttpRange, HttpSearchResultItem, NodeItem, PackedRTree};
 use crate::properties_reader::FgbFeature;
 use crate::{check_magic_bytes, HEADER_MAX_BUFFER_SIZE};
 use crate::{Error, Result};
 use byteorder::{ByteOrder, LittleEndian};
-use bytes::{BufMut, BytesMut};
+use bytes::{BufMut, Bytes, BytesMut};
 use http_range_client::BufferedHttpRangeClient;
+use std::ops::Range;
+
+// The largest request we'll speculatively make.
+// If a single huge feature requires, we'll necessarily exceed this limit.
+const DEFAULT_HTTP_FETCH_SIZE: usize = 1_048_576; // 1MB
 
 /// FlatGeobuf dataset HTTP reader
 pub struct HttpFgbReader {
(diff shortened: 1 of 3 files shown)


src/rust/src/http_reader.rs
```diff
@@ -206,6 +197,165 @@ impl AsyncFeatureIter {
...
+            let wasted_bytes = search_result_item.range.start() - prev_end;
+            if wasted_bytes < combine_request_threshold {
+                if wasted_bytes == 0 {
+                    trace!("adjacent feature");
+                } else {
+                    trace!("wasting {wasted_bytes} to avoid an extra request");
+                }
+                latest_batch.push(search_result_item.range)
+            } else {
+                debug!("creating a new request for batch rather than wasting {wasted_bytes} bytes");
```
pka, line 311: trace level`?

src/rust/src/packed_r_tree.rs
```diff
@@ -188,7 +188,7 @@ async fn read_http_node_items(
 pub struct SearchResultItem {
     /// Byte offset in feature data section
     pub offset: usize,
-    /// Feature number
+    /// Feature number (REVIEW: why do we need this?)
```
pka, line 191: Checked it, it's only used in packed_r_tree tests. It stems most probably from a direct translation of the original C++ implementation.

src/rust/src/packed_r_tree.rs
```diff
@@ -582,42 +587,61 @@ impl PackedRTree {
...
+                    let start = feature_begin + node_item.offset as usize;
+                    if let Some(next_node_item) = &node_items.get(node_pos + 1) {
+                        let end = feature_begin + next_node_item.offset as usize;
+                        results.push(HttpSearchResultItem {
+                            range: HttpRange::Range(start..end),
+                            index: node_id,
+                        });
+                    } else {
+                        debug_assert_eq!(node_pos, num_items);
+                        debug!("No next length (final item)");
```
pka, line 626: still needed?

pka (approved):

PR #322, opened 2023-10-08
[go] Add option to store features in memory during write

### LINKED ISSUE #321: (Go) Usage of temp file in writer.Write
The Go implementation uses a temp file to store the generated features temporarily:
https://github.com/flatgeobuf/flatgeobuf/blob/f25bfdeccace2abc7b65b7fb104a50563a2c1a76/src/go/writer/writer.go#L93-L94

I wonder why a temp file is used instead of an in-memory buffer and if the project is open to not using a temp file.

My motivation is that I want to compile a code that uses FlatGeobuf's Go package into WebAssembly (WASI). However, since using temp files means access to the host file system, it is not easy (I can't find a way yet) to do so.

shihanng:
A proposal based on the discussion in https://github.com/flatgeobuf/flatgeobuf/issues/321

Close #321


bjornharrtell (approved):

bjornharrtell:
@bruno-ga are you ok with this?

bruno-ga:
> @bruno-ga are you ok with this?

Looking.

src/go/writer/writer.go
```diff
@@ -34,19 +37,39 @@ func (f *featureItem) NodeItem() index.NodeItem {
...
+		w.useMemory = true
+	}
+}
+
 // NewWriter returns a new writer instance that will write a flatgeobuf file
 // with the given Header, a possible Index (depending on includeIndex), a
 // FeatureGenerator that will provide the Features to be written and a
 // HeaderUpdater that will be used to update the Header after all features
 // have been generated.
 func NewWriter(header *Header, includeIndex bool,
```
bruno-ga, line 57: Can you please add a new constructor like function instead? It will be better if NewWriter keeps working as before as this is a breaking change to the current API and it is easy enough to add a new one. Something like:

func NewWriterWithOptions(featureGenerator FeatureGenerator, headerUpdater HeaderUpdater, opts ...writerOption) *Writer {
  // Your implementation.
}

func NewWriter(featureGenerator FeatureGenerator, headerUpdater HeaderUpdater) *Writer {
  return NewWriterWithOptions(featureGenerator, headerUpdater, nil)
}

src/go/writer/writer.go
```diff
@@ -34,19 +37,39 @@ func (f *featureItem) NodeItem() index.NodeItem {
 	return *f.nodeItem
 }
 
+// writerOption is the additional option to be passed into Writer.
+type writerOption func(*Writer)
+
+// Use memory instead of temporary file when generating index
+// or execute HeaderUpdater.
+// Effective only when includeIndex is true or when HeaderUpdater is provided.
```
bruno-ga, line 45: Maybe expand the comment with a warning that this can use arbitrarily large amounts of memory? fgb files can be *VERY* large. Maybe you even want to add some protection based on the amount of available memory?

bruno-ga (changes requested):

src/go/writer/writer.go
shihanng: @bruno-ga thanks for the review. 

The main reason that I am using this option pattern `opts ...writerOption` is to avoid breaking change. As shown in L104 of src/go/flatgeobuf_test.go in PR, the user can call the NewWriter function without passing `writer.WithMemory()` and retain the original behavior.

Based on the above, do you still think we need another construtor?

bruno-ga: Sorry, you are correct and this is an aspect of go variadic functions that I have not explored recently and kinda forgot about it. The parameter can, in fact, be omitted so this is fine.

src/go/writer/writer.go
shihanng: > Maybe expand the comment with a warning that this can use arbitrarily large amounts of memory?

Ah good point. I will refine the comment.

> Maybe you even want to add some protection based on the amount of available memory?

I think this will require us to use the `runtime` library, which might impact the performance of the flatgeobuf library. I think we can start with a warning message in the comment. We can add the protection later if it is really necessary. WDYT?

shihanng:
@bruno-ga @bjornharrtell Thank you for the reviews 🙏 .

I added one last commit 0cd4746d2f31fc6258c4da7fcd970608c3cd8553 to warn the caller about memory usage.

If you agree with my comment in https://github.com/flatgeobuf/flatgeobuf/pull/322#discussion_r1358237386 , this is ready to merge.

src/go/writer/writer.go
bruno-ga: Works for me.

bruno-ga:
> @bruno-ga @bjornharrtell Thank you for the reviews 🙏 .
> 
> I added one last commit [0cd4746](https://github.com/flatgeobuf/flatgeobuf/commit/0cd4746d2f31fc6258c4da7fcd970608c3cd8553) to warn the caller about memory usage.
> 
> If you agree with my comment in [#322 (comment)](https://github.com/flatgeobuf/flatgeobuf/pull/322#discussion_r1358237386) , this is ready to merge.

Go for it.

PR #325, opened 2023-10-17
Support of Z and M in Java binding

nicolas-f:
The PR add the support of Z and M in Java binding.

Some unit test are added

Thanks

nb: We will add support for FlatGeobuf in the H2GIS library.


bjornharrtell (approved):
Thanks @nicolas-f this looks nice.

PR #328, opened 2023-10-27
[rust] CI clears cache when Cargo.toml changes

michaelkirk:
Previously it was looking for Cargo.lock which we do not version control.

Note that because we do not commit Cargo.lock, that this approach will gradually pull in the latest dependencies as we rebuild.

If we want all dependencies to be built from cache, we'd need to commit a Cargo.lock. The downside to that approach is we might miss a breakage with a newer dependency release - one that users are likely to be using and that we should probably know about.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -115,7 +115,7 @@ jobs:
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
             src/rust/target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+          key: ${{ runner.os }}-cargo-${{ hashFiles('src/rust/Cargo.toml') }}
           restore-keys: ${{ runner.os }}-cargo-
       - run: cd src/rust && cargo test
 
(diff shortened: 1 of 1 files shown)


michaelkirk:
You can see the first run here:
https://github.com/flatgeobuf/flatgeobuf/actions/runs/6665841911/job/18116251764#step:33:5
```
...
Cache Size: ~492 MB (516029266 B)
Cache saved successfully
Cache saved with key: Linux-cargo-b182ef64dddc9f47af4cb9e96aab8112aef731b831f9a8dbfb0424cf87206328
...
```

And then I re-ran it to verify caching was working: https://github.com/flatgeobuf/flatgeobuf/actions/runs/6665841911/job/18116867668#step:5:27
```
...
Cache restored successfully
Cache restored from key: Linux-cargo-b182ef64dddc9f47af4cb9e96aab8112aef731b831f9a8dbfb0424cf87206328
...
```

bjornharrtell:
Hmm I think we should go with this. (and accept that it could break due to unexpected dependency change)

Issue #329, opened 2023-11-02
Uncaught (in promise) RangeError: Offset is outside the bounds of the DataView


yonghah:
After I update to 3.27 from 3.24 in JS, I am often getting an error like below:

```
packedrtree.ts:179  Uncaught (in promise) RangeError: Offset is outside the bounds of the DataView
    at DataView.getBigUint64 (<anonymous>)
    at packedrtree.ts:179:1
    at streamSearch (packedrtree.ts:189:1)
    at streamSearch.next (<anonymous>)
    at HttpReader.selectBbox (http-reader.ts:139:1)
    at async deserializeFiltered (featurecollection.ts:131:1)
    at async loadFGB (loadFeaturesFromFGB.js:18:1)
    at async RoiControls.js:23:1

```
The JS part is nothing special and no change from 3.24

```
const results = [];
  for await (var fgb_src of fgb_srcs) {
    const iter = geojson.deserialize(
      fgb_src, bounds,
      (meta) => { console.log(meta); });
    for await (let f of iter) {
      results.push(f);
    }
    setCount(results.length);
  }
```
It seems that it is related to the recent change made in 
https://github.com/flatgeobuf/flatgeobuf/commit/cb12470a30441119472d24fb74b08b27b81aaea1

Are there any other getting the same error? It happens quite often. But I'm not able to reproduce precisely.

bjornharrtell:
Thanks @yonghah for reporting this seems unexpected. I might consider rolling back but would really like to reproduce and understand who this could happen.

michaelkirk:
Is the data something you can share? Would you be able to add some logging so you can see which bounding box query exhibits the failed behavior?

yonghah:
@bjornharrtell @michaelkirk 

The dataset is
https://hd20-hlnjc-human-l2.s3.amazonaws.com/asset/sge/full_sdge.fgb

And the boundary is:
{maxX: 98245.28445015292, maxY: 1030.597977715265, minX: 98124.34087916819, minY: 900.6956236546853}

Both are in EPSG:3857.  This is the metadata from deserialize function:
```
{
    "geometryType": 1,
    "columns": [
        {
            "name": "gene_name",
            "type": 11,
            "title": null,
            "description": null,
            "width": 0,
            "precision": -1,
            "scale": -1,
            "nullable": true,
            "unique": false,
            "primary_key": false
        },
        {
            "name": "cnt_total",
            "type": 3,
            "title": null,
[...]

PR #330, opened 2023-11-07
Change bounds check for featureLength calc

bjornharrtell:
This seems to fix https://github.com/flatgeobuf/flatgeobuf/issues/329 but I'm not sure of all the why and consequences of it. Something seems off comparing pos to numItems.

I'm afraid I never quite understood this logic @michaelkirk.

### DIFF
--- a/src/ts/packedrtree.ts
+++ b/src/ts/packedrtree.ts
@@ -171,7 +171,7 @@ export async function* streamSearch(
 
             if (isLeafNode) {
                 const featureLength = (() => {
-                    if (pos < numItems - 1) {
+                    if (pos < end - 1) {
                         // Since features are tightly packed, we infer the
                         // length of _this_ feature by measuring to the _next_
                         // feature's start.


bjornharrtell:
I'm now not actually sure featureLength is used or if it was just thought as an extra nice to have info for the caller? In either case I'm starting to believe this must have been wrong before too and that it went out of bounds but that is not fatal for typed array index access.

src/ts/packedrtree.ts
```diff
@@ -171,7 +171,7 @@ export async function* streamSearch(
 
             if (isLeafNode) {
                 const featureLength = (() => {
-                    if (pos < numItems - 1) {
+                    if (pos < end - 1) {
```
michaelkirk, line 174: I don't think this change is correct. 

Here's my understanding:

`featureLength` the length, in bytes, of this feature in the feature buffer.

`numItems`: the number of leaf nodes in the tree. AKA the number of features in the dataset.

`pos`: the index of the feature in the feature set, which is the same as the ordering of the leaf nodes. (aka it's hilbert ordering relative to the other features in the feature set).

`end`: the index of the last `byte` in the batch of index nodes we're considering.

After perusing the index to figure out where our features live, we fetch the features from the feature buffer. For transfer efficiency and speed, we want to fetch "just enough" data.  Knowing each feature's `featureLength` is how we do that. 

With the PackedRTree index we can figure out the `featureLength` of every feature *except* for the very last feature in the entire dataset. That's what this branch is about - hence the comparison to `numItems`. 

It might be clearer to invert the condition and write it as:

```
console.assert(pos <= numItems, "can't have an index beyond the number of features");
const isLastFeatureInDataset = pos == end;
if (isLastFeatureInDataset) {
  // This is the last feature - there's no "next" feature
  // to measure to, so we can't know it's length.
  return null;
} else {
  // Since features are tightly packed, we infer the
  // length of _this_ feature by measuring to the _next_
  // feature's start.
  const nextPos = (pos - nodeIndex + 1) * 40;
  const nextOffset = dataView.getBigUint64(
    nextPos + 32,
    true,
  );
  return nextOffset - offset;
}
```

So anyway, I hope it's clear now it doesn't make sense to compare `pos` (the index of a feature) to `end` the index of a byte.

I'll take a look and see if I can get to the bottom of things... and maybe make a PR to clear up some of these variable names, which are currently not very helpful.

bjornharrtell: Ok but as the buffer (that the dataView spans) is, depending on size etc, possibly only a subset of the nodes in a tree level it also doesn't seem right it can always peek the nextOffset up to the last feature. (i.e that would only be true of the full set of nodes is in the buffer?)

michaelkirk:
Superseded by #331

PR #331, opened 2023-11-08
[ts] fix OOB error

### LINKED ISSUE #329: Uncaught (in promise) RangeError: Offset is outside the bounds of the DataView

michaelkirk:
Fixes #329 

@bjornharrtell, you were on the right track when you said:
> it also doesn't seem right it can always peek the nextOffset up to the last feature. (i.e that would only be true of the full set of nodes is in the buffer?)

A little more precisely, we don't need *all* the nodes, just one extra node past our current node range. I accounted for this when I did the rust version here:
https://github.com/flatgeobuf/flatgeobuf/pull/319/files#diff-4735ecfc66255497565c8404679ce4037b304a96fe24f7a9aea016c224be6110R597 [^see footnote]

...but I failed to account for the "get an extra node" logic when I wrote the javascript version 😞. When I wrote it, the consequences of my bug were relatively modest - we weren't optimally sizing the feature fetch for all of our feature requests, but since we ultimately defer to the feature_len read from the feature buffer, we'd still get the data, but it might require an extra fetch like in the olden days. 

With the introduction of cb12470, the consequences become more severe, because the existing bug now resulted in an OOB exception. Sorry for setting a trap for you @bjornharrtell!

The actual fix is small, in d8512b80ecc248035fa68fb82a39b5cada20f9ce

But I also included some preparatory commits that helped me understand the code well enough to fix it. I tried to have my commits well factored if you'd like to review them individually, but if you prefer, I can close this PR and open up multiple smaller PRs.

[^footnote] 
There was actually a bug here in the rust version too, albeit a smaller one. The rust code was comparing to the wrong variable, and thus being overly conservative about guessing feature size. Fixed in commit:`[rust] Fix when to fetch next node`

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "@vitest/coverage-v8": "^0.34.6",
     "eslint": "8.51.0",
     "jsts": "2.11.0",
+    "local-web-server": "^5.3.0",
     "node-fetch": "^3.3.2",
     "ol": "8.1.0",
     "prettier": "3.0.3",
--- a/src/rust/src/packed_r_tree.rs
+++ b/src/rust/src/packed_r_tree.rs
@@ -278,9 +278,8 @@ pub fn calc_extent(nodes: &[NodeItem]) -> NodeItem {
 pub struct PackedRTree {
     extent: NodeItem,
     node_items: Vec<NodeItem>,
-    num_items: usize,
-    num_nodes: usize,
-    node_size: u16,
+    num_leaf_nodes: usize,
+    branching_factor: u16,
     level_bounds: Vec<(usize, usize)>,
 }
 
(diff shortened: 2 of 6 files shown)


bjornharrtell:
Very nicely done @michaelkirk, thanks! I'm fine with a rebase merge of this so I will do that now.

PR #337, opened 2023-11-18
[rust] Clearer code for HttpReader

michaelkirk:
I'm not sure if people are interested in this kind of non-functional change. No hard feelings either way.

### DIFF
--- a/src/rust/src/packed_r_tree.rs
+++ b/src/rust/src/packed_r_tree.rs
@@ -10,6 +10,7 @@ use std::cmp::{max, min};
 use std::collections::VecDeque;
 use std::io::{Cursor, Read, Seek, SeekFrom, Write};
 use std::mem::size_of;
+use std::ops::Range;
 
 #[derive(Clone, PartialEq, Debug)]
 #[repr(C)]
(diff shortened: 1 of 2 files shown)


src/rust/src/packed_r_tree.rs
```diff
@@ -587,44 +584,16 @@ impl PackedRTree {
...
-            let start_node = next.nodes.start;
-            let is_leaf_node = next.level == 0;
-            if is_leaf_node {
-                assert!(start_node >= leaf_nodes_offset);
-            } else {
-                assert!(start_node < leaf_nodes_offset);
-                assert!(next.nodes.end < leaf_nodes_offset);
-            }
-            // find the end index of the nodes
-            let mut end_node = min(
```
michaelkirk, line 604: All this index offsetting now happens *before* the node ranges are enqueued rather than cleaning up after the fact.

src/rust/src/packed_r_tree.rs
```diff
@@ -637,45 +606,69 @@ impl PackedRTree {
...
                         });
                     }
-                    continue;
-                }
-
-                let combine_request_node_threshold =
-                    combine_request_threshold / size_of::<NodeItem>();
-
-                // Add node to search recursion
-                match (queue.back_mut(), node_item.offset as usize) {
```
michaelkirk, line 647: I broke this complex match statement into a series of dumber if blocks. It's not a little more redundant, but I think easier to follow.

bjornharrtell (approved):
I'm all for making code more clear, especially when it also reduces the net amount of code :)

michaelkirk:
I snuck in one more commit to fix an off-by-one in the debug assertion. Please take another look @bjornharrtell and/or @pka

michaelkirk:
I'm going to be bold and assume fixing a debug assert and adding a unit test are not controversial.

PR #339, opened 2023-12-21
Fix HTTP 416 on some webservers

michaelkirk:
- [ts] NBD: clarify logging message
- [ts] NBD: use non-deprecated spelling
- [rust] Fix HTTP 416 on some webservers
- [ts] Fix HTTP 416 on some webservers

### DIFF
--- a/src/rust/src/http_reader.rs
+++ b/src/rust/src/http_reader.rs
@@ -324,14 +324,20 @@ impl FeatureBatch {
         let last = feature_ranges
             .last()
             .expect("We never create empty batches");
-        let covering_range = first.clone().with_end(last.end());
-        let min_request_size = match covering_range.length() {
-            // Should only happen for the final feature batch.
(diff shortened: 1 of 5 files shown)


bjornharrtell:
Turns out it's rather difficult/involved to fully reproduce the situation that has been observed. Wanted to reproduce and try this fix there but as of yet unable to.

bjornharrtell (approved):
Done some testing that indicates this is looking good.

michaelkirk:
You shouldbe able to reproduce it every time by fetching a bounding box that covers the last feature of the dataset.To figure that out, I did “select all” and logged the geometry of the last item (compare idx to the feature count from the header). On Dec 23, 2023, at 06:11, Björn Harrtell ***@***.***> wrote:﻿
Merged #339 into master.

bjornharrtell:
Looks like it was reproduced by an empty bbox too :D - not sure that ever was intended to work at all though...

PR #344, opened 2024-02-02
[net] Avoid reading spatial index when seeking is available

bjornharrtell:


bjornharrtell:
Hi @FObermaier, this may be of interest to you. I found myself needing to read FlatGeobuf through HTTP with spatial filtering and I noticed that it read the full spatial index when it should be able to use seek. This should be significant also for block storage streams but of course for network streams it's of extreme importance. The fix was seemingly very simple.

PS. I use https://github.com/espresso3389/HttpStream which implements seekable streams for HTTP with range requests.

PR #345, opened 2024-02-03
Rust: Prototype of generic HttpFgbReader

kylebarron:
Right now the Flatgeobuf crates has a synchronous `FgbReader` and an async `HttpFgbReader`. But the `HttpFgbReader` is not generic. It _requires_ an underlying `reqwest::Client`. It would be great to allow this underlying backend to be generic.

The motivation for this is to connect Flatgeobuf to the [object-store crate](https://docs.rs/object_store/latest/object_store/), which abstracts among various object storage providers including S3, GCS, Azure, etc. In https://github.com/geoarrow/geoarrow-rs/pull/494 I have an example of connecting object-store to `http_range_client::AsyncHttpRangeClient` (https://github.com/geoarrow/geoarrow-rs/pull/494/files#diff-4bc105a5ea08ab4e7e6d0bea2453dcd5e3a7c89d531019042abfeeca30b4b83c), but I can't construct a `FlatGeobuf::HttpFgbReader` that's generic over that backend.

### DIFF
--- a/src/rust/Cargo.toml
+++ b/src/rust/Cargo.toml
@@ -13,7 +13,7 @@ keywords = ["geo", "r-tree", "spatial"]
 
 [features]
 default = ["http"]
-http = ["http-range-client", "bytes"]
+http = ["http-range-client", "bytes", "reqwest"]
 
 [dependencies]
 flatbuffers = "23.5.26"
@@ -24,6 +24,7 @@ bytes = { version = "1.5.0", optional = true }
 log = "0.4.20"
 fallible-streaming-iterator = "0.1.9"
 tempfile = "3.8.1"
+reqwest = { version = "0.11.22", optional = true }
 
 [dev-dependencies]
 geozero = { version = "0.11.0", default-features = true }
--- a/src/rust/src/http_reader.rs
+++ b/src/rust/src/http_reader.rs
@@ -47,14 +47,6 @@ impl<T: AsyncHttpRangeClient> HttpFgbReader<T> {
         Self::_open(client).await
     }
 
-    // pub async fn open(url: &str) -> Result<HttpFgbReader<reqwest::Client>> {
-    //     trace!("starting: opening http reader, reading header");
-    //     let mut client = BufferedHttpRangeClient::new(url);
-    //     let x = Self::_open(client).await?;
-    //     Ok(Self::_open(client).await?)
-
-    // }
-
     async fn _open(mut client: AsyncBufferedHttpRangeClient<T>) -> Result<HttpFgbReader<T>> {
         // Because we use a buffered HTTP reader, anything extra we fetch here can
         // be utilized to skip subsequent fetches.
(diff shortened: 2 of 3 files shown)


michaelkirk:
Hi @kylebarron - is there a reason that this is opened as a draft? What further changes are you intending to make?

kylebarron:
> Hi @kylebarron - is there a reason that this is opened as a draft? What further changes are you intending to make?

Mostly just that it was a proposal and I hadn't opened an issue previously to discuss it, and I wasn't sure if people were interested in such a change.

michaelkirk (approved):
The goal seems reasonable and the changes LGTM. 

As well as this flatgeobuf crate, @pka also maintains the `http-range-client` crate, so it'd be good to get a review from him as well.

pka (commented):
I'm all for supporting other HTTP clients.

My initial idea was to include other clients in the `http-range-client` crate and build with approprate feature flags. But having the possibilty to use your own client implementation with flatgeobuf would justify having a generic `HttpFgbReader`.

src/rust/Cargo.toml
```diff
@@ -13,7 +13,7 @@ keywords = ["geo", "r-tree", "spatial"]
 
 [features]
 default = ["http"]
-http = ["http-range-client", "bytes"]
+http = ["http-range-client", "bytes", "reqwest"]
```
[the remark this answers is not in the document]
kylebarron: I think this was to keep backwards compatibility. Reqwest is already a dependency via `http-range-client`, but now we need to import the `reqwest::Client` type to keep `HttpFgbReader::open` the same.

src/rust/Cargo.toml
```diff
@@ -24,6 +24,7 @@ bytes = { version = "1.5.0", optional = true }
 log = "0.4.20"
 fallible-streaming-iterator = "0.1.9"
 tempfile = "3.8.1"
+reqwest = { version = "0.11.22", optional = true }
```
kylebarron, line 27: It might be preferable to relax the pin here to `0.11`? 0.11.22 is latest. Or maybe that doesn't matter. http-range-client is set to 0.11:
https://github.com/pka/http-range-client/blob/19d25265b9172fdbdd1d2b5dfd095e90c22c2408/Cargo.toml#L24
```suggestion
reqwest = { version = "0.11", optional = true }
```

michaelkirk: It's preferable to use the oldest version you have actually tested and verified works. In practice this is usually the latest version when developing. There's marginal utility in supporting older versions. 

Note that in rust,version pinning is different from some other versioning schemes and will use the latest semver compatible with this notation. 

https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277/1

kylebarron:
Can I do something to help move this PR forward? Anything blocking merge?

PR #351, opened 2024-03-09
Upgrade dependencies

bjornharrtell:


ianthetechie:
I'd love to see this move forward if possible :) In particular, upgrading `reqwest` will be nice to have since it's built on `hyper` 1.x. Any crates that have upgraded either `hyper` or `reqwest` currently need to build multiple versions, which isn't ideal.

bjornharrtell:
@ianthetechie rust deps upgrade done in https://github.com/flatgeobuf/flatgeobuf/pull/378, I will revive at some point.

PR #355, opened 2024-03-22
[go] fix geometry parts build

durandAD:
First thank you for this great go package. When using this package I had a few issues while trying to write multipolygons and using parts of geometry.

I had 2 issues that are fixed by this PR.
- When writing the geometry that contains parts, the build always caused a panic:
> panic: Incorrect creation order: object must not be nested.
- When providing only parts of a geometry the geometry was not indexed as the parts where not used to create it. 

I added a test to show what kind of feature was not working before

### DIFF
--- a/src/go/flatgeobuf_test.go
+++ b/src/go/flatgeobuf_test.go
@@ -77,9 +77,6 @@ func TestCreateFGBFileAndBasicSearch(t *testing.T) {
 				createSquareFeature(-1, 0, 0, 1, 2),
 				createSquareFeature(-1, -1, 0, 0, 3),
 				createSquareFeature(0, -1, 1, 0, 4),
-				createMultiPolygonFeature([][]float64{
-					{4, 4, 5, 5}, {5, 5, 6, 6}, {10, 4, 11, 5},
-				}, 5),
 			},
 		}
 	}
@@ -93,7 +90,7 @@ func TestCreateFGBFileAndBasicSearch(t *testing.T) {
 		header := writer.NewHeader(headerBuilder).
 			SetName("Households ShapeFile Data").
 			SetTitle("Households ShapeFile Data").
-			SetGeometryType(flattypes.GeometryTypeUnknown)
+			SetGeometryType(flattypes.GeometryTypePolygon)
 		householdsCol := writer.NewColumn(headerBuilder).
 			SetName("Households").
 			SetType(flattypes.ColumnTypeUInt)
@@ -105,7 +102,9 @@ func TestCreateFGBFileAndBasicSearch(t *testing.T) {
 
 	var mockFile bytes.Buffer
 	wr := writer.NewWriter(header, true, fgen(), hu)
-	wr.Write(&mockFile)
+	if _, err := wr.Write(&mockFile); err != nil {
+		t.Errorf("failed to write in buffer: %v", err)
+	}
 
 	// TEST:
 	// check header metadata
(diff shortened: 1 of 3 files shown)


bruno-ga:
Thanks for your PR. I will try to review it by EOD today.

bjornharrtell:
LGTM. @bruno-ga you ok with merge?

src/go/flatgeobuf_test.go
```diff
@@ -77,6 +77,9 @@ func TestCreateFGBFileAndBasicSearch(t *testing.T) {
 				createSquareFeature(-1, 0, 0, 1, 2),
 				createSquareFeature(-1, -1, 0, 0, 3),
 				createSquareFeature(0, -1, 1, 0, 4),
+				createMultiPolygonFeature([][]float64{
```
bruno-ga, line 80: Would you mind moving this to its own test? It seems important enough to deserve this (also it might not match the basic test here).

src/go/flatgeobuf_test.go
```diff
@@ -90,7 +93,7 @@ func TestCreateFGBFileAndBasicSearch(t *testing.T) {
 		header := writer.NewHeader(headerBuilder).
 			SetName("Households ShapeFile Data").
 			SetTitle("Households ShapeFile Data").
-			SetGeometryType(flattypes.GeometryTypePolygon)
+			SetGeometryType(flattypes.GeometryTypeUnknown)
```
bruno-ga, line 96: If you move the changes to their own test, you do not need to use Unknown here (which again, goes for the "basic" part of this test). :)

src/go/flatgeobuf_test.go
```diff
@@ -137,6 +140,8 @@ func TestCreateFGBFileAndBasicSearch(t *testing.T) {
 			expectedResultProperties: []int{1, 2, 3, 4}},
 		{searchMinX: 2, searchMinY: 2, searchMaxX: 3, searchMaxY: 3,
 			expectedResultProperties: []int{}},
+		{searchMinX: 4.5, searchMinY: 4.5, searchMaxX: 4.6, searchMaxY: 4.6,
```
bruno-ga, line 143: Ditto. More to own test.

src/go/writer/geometry_extended.go
```diff
@@ -13,14 +13,40 @@ func NewGeometryExtended(geometry *Geometry) *GeometryExtended {
...
+func geomBoundingBox(g *Geometry) (minX, minY, maxX, maxY float64) {
 	firstPartEnd := len(g.xy)
 	if firstPartEnd == 0 {
 		// Nothing to do.
 		return
 	}
 
 	if len(g.ends) != 0 {
-		firstPartEnd = int(g.ends[0])
+		firstPartEnd = int(g.ends[0]) * 2
```
bruno-ga, line 49: Can you add a comment here to explain why you are multiplying by 2? It is not obvious form the code around it and it is also not clear to me why you need to do it (although I might be missing something).

bruno-ga (changes requested):
Looks good generally. Just a few changes suggested to make this a bit clearer.

bruno-ga:
> LGTM. @bruno-ga you ok with merge?

Sorry, I punted reviewing it and then forgot about it. Just did it now. Will wait for the author's  reply.

@bjornharrtell BTW, heads up that I will start using my personal account to work on this, @brunoga (no dash).

bjornharrtell:
@brunoga thanks for your input and contribution. :)

brunoga:
> @brunoga thanks for your input and contribution. :)

Thanks. :)

src/go/flatgeobuf_test.go
durandAD: Good point, I created another test

src/go/writer/geometry_extended.go
durandAD: I added a comment. don't hesitate to tell me if it is not clear enough

src/go/flatgeobuf_test.go
brunoga: Thanks.

src/go/writer/geometry_extended.go
brunoga: LGTM.

bruno-ga (approved):

brunoga:
@bjornharrtell Apparently I can not approve with either account so up to you. :)

bjornharrtell (approved):

bjornharrtell:
Thanks for your contribution @durandAD!

Issue #358, opened 2024-04-08
[ts] Timeout when loading fgb files using takeAsync+deserialize starting with v3.27.3


twelch:
When using flatgeobuf Javascript package version `3.27.3` and higher in a node v20.12.1 environment, I do not receive a response when loading a flatgeobuf from an external URL using deserialize + takeAsync.

I can't tell exactly what is happening, it's not throwing an error for me, but I have reproduced the issue - https://github.com/twelch/flatgeobuf-error-demo/tree/main.  see README for steps, the test (which is based on this [internal fgb test](https://github.com/flatgeobuf/flatgeobuf/blob/2828e07fd759aa8b44c6b1aaab41320b22dd0d29/src/ts/geojson.spec.ts#L393)) will timeout as-is and after downgrading to `3.27.2` it will pass.

This seems to only occur when I am using the published library, adding this test to the internal flatgeobuf suite and it will pass/load just fine. 
This seems to occur with any flatgeobuf loaded using this method, at least all of mine, not just this one published fgb, which is a simple polygon boundary for the world.

```json
{
    "type": "FeatureCollection",
    "features": [
      {
        "type": "Feature",
        "properties": {
          "name": "World boundary",
          "description": "World"
        },
        "geometry": {
          "coordinates": [
            [
              [
                -180,
                90
              ],
              [
                -180,
                -90
              ],
              [
                180,
                -90
              ],
              [
                180,
                90
              ],
              [
                -180,
                90
              ]
            ]
          ],
          "type": "Polygon"
        }
      }
    ]
  }
```

bjornharrtell:
Thanks for the report and nice reproduction. I have noticed something similar and found it it happens only on node native fetch api. The tests in flatgeobuf works because it includes node-fetch polyfill (globally). I have as of yet no clue why it times out with node native fetch which is still marked as experimental.

Because node native fetch is experimental I'm not sure how deep I want to dig at this time. That said, it's a nice find that it worked in 3.27.2, that means it likely has some relation to dependency upgrades done in https://github.com/flatgeobuf/flatgeobuf/commit/987499b2abf4bab8b070529c4d5228b62992b843 and https://github.com/flatgeobuf/flatgeobuf/commit/6c0124cd25586d483530c113ada4bcc82e38ffef. Main suspect is web-streams-polyfill.

twelch:
That's helpful. I can confirm that polyfilling with node-fetch in the test file gets it to pass as you said.

[...]

PR #359, opened 2024-04-08
[ts] Use native node fetch removing polyfills

bjornharrtell:


bjornharrtell:
This isn't working due to https://github.com/flatgeobuf/flatgeobuf/issues/358.

Issue #361, opened 2024-04-16
Error: IO(Error { kind: UnexpectedEof, message: "failed to fill whole buffer" })


lixiaofei123:
Hello, I'm using flatgeobuf in Rust to read fgb files. Sometimes, when specifying select_bbox, I encounter the following error:
Error: IO(Error { kind: UnexpectedEof, message: "failed to fill whole buffer" })

The error does not occur consistently, it only happens when certain values are passed as parameters to select_bbox.

I tried creating fgb files using the gdaltindex command or FgbWriter separately, but encountered the same error.

My code is as follows:

```rust
fn main() -> Result<(), Box<dyn Error>> {
    let fgbpath = "countries.fgb";
    let file = fs::File::open(&fgbpath)?;
    let mut filein = BufReader::new(file);
    let fgb = FgbReader::open(&mut filein)?;

    let mut features = fgb
        .select_bbox(13034882.870499391, 3741363.2234745212, 13034959.307527676, 3741439.6605028082)?;

    while let Some(feature) = features.next()? {
        println!("{}", feature.to_json()?);
    }

    Ok(())
}
```

The versions of the related dependencies I'm using are:

```toml
flatgeobuf = "4.1.0"
geozero = "0.12.0"
```

The fgb file I generated is below.

[countries.zip](https://github.com/flatgeobuf/flatgeobuf/files/14988041/countries.zip)

lixiaofei123:
> Hello, I'm using flatgeobuf in Rust to read fgb files. Sometimes, when specifying select_bbox, I encounter the following error: Error: IO(Error { kind: UnexpectedEof, message: "failed to fill whole buffer" })
> 
> The error does not occur consistently, it only happens when certain values are passed as parameters to select_bbox.
> 
> I tried creating fgb files using the gdaltindex command or FgbWriter separately, but encountered the same error.
> 
> My code is as follows:
> 
> ```rust
> fn main() -> Result<(), Box<dyn Error>> {
>     let fgbpath = "countries.fgb";
>     let file = fs::File::open(&fgbpath)?;
>     let mut filein = BufReader::new(file);
>     let fgb = FgbReader::open(&mut filein)?;
> 
>     let mut features = fgb
>         .select_bbox(13034882.870499391, 3741363.2234745212, 13034959.307527676, 3741439.6605028082)?;
> 
>     while let Some(feature) = features.next()? {
>         println!("{}", feature.to_json()?);
>     }
> 
>     Ok(())
[...]

PR #362, opened 2024-04-18
[rust] Fix "UnexpectedEof" error when bbox results includes first item 

michaelkirk:
Note: The error doesn't occur until reading the *next* item.

tldr; we had some bad accounting for tracking the file position that only affected bbox calculations that covered the first feature in the buffer + at least one other feature.

## What's going on here

We want to set `count` in the iter initializer.

FGB includes `count` in the header, but it's optional. Because flatbuffer doesn't have null types, we can't distinguish between `count` being unset and `count` truly being zero.

So the way we detect an "empty" FGB is to try to read from the feature buffer in the iter's initializer. If the feature buffer is empty, this read will fail and we can immediately mark the FGB reader as "finished", which allows the iter to be well behaved elsewhere.

The other factor is that, as we read features, we track the offset into the feature buffer (`cur_pos`) so that we know how far we need to Seek from where we currently are when jumping to the next feature in the BBOX
results.

So, every time we read a feature_size, we've read four bytes and then when we read the feature we've moved whatever the feature_len encoded in those four bytes.

**BUT THE THING IS** - we weren't updating the cur_pos immediately after
reading feature_len, we updated it after reading the feature, and
included an extra 4 bytes at that point.

That usually works.

However - in the case of the very first feature, we've already added the
"4" to the feature offset. That's a sentinel value we check for to infer
it's the first feature so that we don't read the feature_len *again*.

So in the case that the bbox included the first feature, we were adding
an *extra* 4 to the cur_pos.

## The fix is

We now advance cur_pos immediately after we read the feature length. After reading the feature, we only add feature_len, not feature_len+4.

I've also replaced the `finished` bool and the `cur_pos == 4` check with a more explicable `state` enum.

Fixes https://github.com/flatgeobuf/flatgeobuf/issues/361

### DIFF
--- a/src/rust/src/file_reader.rs
+++ b/src/rust/src/file_reader.rs
@@ -33,7 +33,7 @@ pub struct FeatureIter<R, S> {
     feat_no: usize,
     /// File offset within feature section
     cur_pos: u64,
-    /// All features read or end of file reached
+    /// Reading state
     state: State,
     /// Whether or not the underlying reader is Seek
     seekable_marker: PhantomData<S>,
--- a/src/rust/src/file_reader.rs
+++ b/src/rust/src/file_reader.rs
@@ -34,9 +34,17 @@ pub struct FeatureIter<R, S> {
     /// File offset within feature section
     cur_pos: u64,
     /// All features read or end of file reached
-    finished: bool,
+    state: State,
     /// Whether or not the underlying reader is Seek
     seekable_marker: PhantomData<S>,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+enum State {
+    Init,
+    ReadFirstFeatureSize,
+    Reading,
+    Finished,
 }
 
 #[doc(hidden)]
@@ -237,6 +245,9 @@ impl<R: Read> FallibleStreamingIterator for FeatureIter<R, NotSeekable> {
         if let Some(filter) = &self.item_filter {
             let item = &filter[self.feat_no];
             if item.offset as u64 > self.cur_pos {
+                if self.state == State::ReadFirstFeatureSize {
+                    self.state = State::Reading;
+                }
                 // skip features
                 let seek_bytes = item.offset as u64 - self.cur_pos;
                 io::copy(&mut (&mut self.reader).take(seek_bytes), &mut io::sink())?;
@@ -289,6 +300,9 @@ impl<R: Read + Seek> FallibleStreamingIterator for FeatureIter<R, Seekable> {
         if let Some(filter) = &self.item_filter {
             let item = &filter[self.feat_no];
             if item.offset as u64 > self.cur_pos {
+                if self.state == State::ReadFirstFeatureSize {
+                    self.state = State::Reading;
+                }
                 // skip features
                 let seek_bytes = item.offset as u64 - self.cur_pos;
                 self.reader.seek(SeekFrom::Current(seek_bytes as i64))?;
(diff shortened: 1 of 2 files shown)


src/rust/src/file_reader.rs
```diff
@@ -421,37 +435,44 @@ mod geozero_api {
...
         };
 
-        FeatureIter {
-            reader,
-            verify,
-            fbs,
-            item_filter,
-            count,
-            feat_no: 0,
-            cur_pos: 4,
```
michaelkirk, line 451: We used to initialize the `cur_pos: 4` here to signify that we've already read the first feature length. This was necessary to correctly compute the offset if we were to seek to a subsequent feature.

src/rust/src/file_reader.rs
```diff
@@ -465,29 +486,46 @@ impl<R: Read, S> FeatureIter<R, S> {
...
                 return true;
             }
         }
         false
     }
 
     /// Read feature size and return true if end of dataset reached
-    fn read_feature_size(fbs: &mut FgbFeature, reader: &mut R) -> bool {
-        fbs.feature_buf.resize(4, 0);
-        reader.read_exact(&mut fbs.feature_buf).is_err()
```
michaelkirk, line 483: Previously we didn't bump the cur_pos when actually reading feature size.

src/rust/src/file_reader.rs
```diff
@@ -497,20 +535,21 @@ impl<R: Read, S> FeatureIter<R, S> {
             let _feature = size_prefixed_root_as_feature(&self.fbs.feature_buf)?;
         }
         self.feat_no += 1;
-        self.cur_pos += self.fbs.feature_buf.len() as u64;
```
michaelkirk, line 500: feature_buf is the size_prefixed_root_as_feature. So it includes the size_prefix which is 4 bytes + feature_size.

In the new code, we *only* add feature_size here since we've already accounted for the +4 bytes closer to the actual reading, in read_feature_size.

src/rust/src/file_reader.rs
```diff
@@ -465,29 +486,46 @@ impl<R: Read, S> FeatureIter<R, S> {
...
-        reader.read_exact(&mut fbs.feature_buf).is_err()
+    fn read_feature_size(&mut self) -> bool {
+        self.fbs.feature_buf.resize(4, 0);
+        self.cur_pos += 4;
+        self.reader.read_exact(&mut self.fbs.feature_buf).is_err()
     }
 
     fn read_feature(&mut self) -> Result<()> {
-        // Read feature size if not already read in select_all or select_bbox
-        if self.cur_pos != 4 && Self::read_feature_size(&mut self.fbs, &mut self.reader) {
```
michaelkirk, line 488: `self.cur_pos != 4` was our sentinel for "This is the first feature and we've already called read_feature_size in the initializer so don't do it again" - now we track an explicit State variable, which I think is a little more explicable.

src/rust/tests/read.rs
```diff
@@ -137,6 +137,23 @@ fn read_bbox() -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn read_bbox_including_first_feature() -> Result<()> {
```
michaelkirk, line 141: If you want to try it out, this test will explode with the old code.

bjornharrtell (approved):
Nice work @michaelkirk - let's see if @pka have time to review, but if not this seems like good to merge to me.

src/rust/src/file_reader.rs
```diff
@@ -34,11 +34,19 @@ pub struct FeatureIter<R, S> {
     /// File offset within feature section
     cur_pos: u64,
     /// All features read or end of file reached
```
pka, line 36: Should be smth. like "Reading state" then.

I feel a temptation to include `cur_pos` in the Reading state, but this would the code probably make less readable.

pka (approved):
I'm sorry for the embarassing implementation error.

The solution contains more changes than expected, but makes absolutely sense.

src/rust/src/file_reader.rs
michaelkirk: > Should be smth. like "Reading state" then.

Whoops good catch!

> I feel a temptation to include cur_pos in the Reading state, but this would probably make the code less readable.

I've attempted this here:
https://github.com/michaelkirk/flatgeobuf/pull/new/mkirk/gh-361-unexpected-eof-cur-pos-in-state

But I think I agree that it's slightly less readable. Or at least not obviously better.

bjornharrtell:
@michaelkirk is this ready to merge?

PR #350, opened 2024-02-23
prepare for rust release

michaelkirk:
I'm not actually a publisher of the crate, but I've prepared a release branch that someone else could publish from.

PR #365, opened 2024-05-28
Prepare rust release 4.2.0

kylebarron:
Based on https://github.com/flatgeobuf/flatgeobuf/pull/350.

It would be nice to cut a new release of rust flatgeobuf so that downstream projects can use flatgeobuf with geozero 0.13


bjornharrtell:
Published crate. Have fun @kylebarron. :)

PR #376, opened 2024-07-23
Fix too-small request sizing after making a large request.

michaelkirk:
Background:

To minimize the number of requests, nearby features are grouped into batches which can be fetched in a single request. While processing the first feature of the batch, we fetch all the data we need for the entire batch.

However, because our HTTP library holds the entire request in memory, to keep memory usage reasonable, there was logic to cap the max size of that http request.

Bug:

If you ever hit that max http request size while fetching a feature batch, each subsequent feature in that batch would require 2 requests per feature — one to read the feature's length and another to read the feature's data. 

This could add up to a lot of requests - see https://github.com/a-b-street/15m/pull/8#issuecomment-2237792900

This faulty logic would fix itself upon starting the subsequent feature batch, until/unless another large request was encountered.

Fix:

We now calculate the size of the batch each request, rather than just once per batch.

I also added a test client to test this and hopefully any other future request related bugs in the future.

### DIFF
--- a/src/rust/Cargo.toml
+++ b/src/rust/Cargo.toml
@@ -25,7 +25,6 @@ log = "0.4.20"
 fallible-streaming-iterator = "0.1.9"
 tempfile = "3.8.1"
 reqwest = { version = "0.11.22", optional = true }
-async-trait = "0.1.81"
 
 [dev-dependencies]
 geozero = { version = "0.13.0", default-features = true }
@@ -38,6 +37,7 @@ tokio = { version = "1.34.0", features = ["rt-multi-thread"] }
 reqwest = { version = "0.11.22", default-features = true }
 geo-types = "0.7.12"
 yocalhost = "0.3.0"
+async-trait = "0.1.81"
 
 [[bench]]
 name = "read"
--- a/src/rust/Cargo.toml
+++ b/src/rust/Cargo.toml
@@ -25,6 +25,7 @@ log = "0.4.20"
 fallible-streaming-iterator = "0.1.9"
 tempfile = "3.8.1"
 reqwest = { version = "0.11.22", optional = true }
+async-trait = "0.1.81"
 
 [dev-dependencies]
 geozero = { version = "0.13.0", default-features = true }
--- a/src/rust/src/http_reader/mock_http_range_client.rs
+++ b/src/rust/src/http_reader/mock_http_range_client.rs
@@ -1,0 +1,105 @@
+use crate::{HttpFgbReader, Result};
+use bytes::Bytes;
+use http_range_client::AsyncHttpRangeClient;
+use std::fs::File;
+use std::io::{BufReader, Read, Seek, SeekFrom};
+use std::ops::Range;
+use std::path::PathBuf;
+use std::sync::{Arc, RwLock};
+
+impl HttpFgbReader<MockHttpRangeClient> {
+    /// NOTE: For debugging expediency, this test class often prefers panics over returning a result.
+    pub async fn mock_from_file(
+        path: &str,
+    ) -> Result<(
+        HttpFgbReader<MockHttpRangeClient>,
+        Arc<RwLock<RequestStats>>,
+    )> {
+        trace!("starting: opening http reader, reading header");
+
+        let stats = Arc::new(RwLock::new(RequestStats::new()));
+        let http_client = MockHttpRangeClient::new(path, stats.clone());
+        let client = http_range_client::AsyncBufferedHttpRangeClient::with(http_client, path);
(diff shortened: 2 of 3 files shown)


michaelkirk:
Sorry, I don't think this is ready to merge yet. The actual http client we use is behaving pretty differently from the test API client. Which is not giving me a lot of confidence in the testing strategy I just introduced. 😞 

I think it might be an interaction between the built in "generic" overfetch strategy of our chosen http library and the more fine tuned overfetch strategy in this library.

michaelkirk:
After a good night's rest, I realized it is in fact fixed!

The thing that confused me was the new logging in our http library. It looks like this:
```
TRACE [read_logger] Read 64778736-64778739 (4 bytes). Total requests: 4579 (7912092 bytes),read,64778736,64778739,4,4,4579,7912092
TRACE [read_logger] Read 64778740-64779295 (556 bytes). Total requests: 4580 (7912648 bytes),read,64778740,64779295,556,556,4580,7912648
TRACE [read_logger] Read 65180376-65180379 (4 bytes). Total requests: 4581 (7912652 bytes),read,65180376,65180379,4,4,4581,7912652
DEBUG [read_logger] Read 65180376-65182287 (1912 bytes). Total requests: 44 (14523928 bytes),http-range,65180376,65182287,1912,4,44,14523928
TRACE [read_logger] Read 65180380-65180895 (516 bytes). Total requests: 4582 (7913168 bytes),read,65180380,65180895,516,516,4582,7913168
TRACE [read_logger] Read 65180896-65180899 (4 bytes). Total requests: 4583 (7913172 bytes),read,65180896,65180899,4,4,4583,7913172
TRACE [read_logger] Read 65180900-65181359 (460 bytes). Total requests: 4584 (7913632 bytes),read,65180900,65181359,460,460,4584,7913632
TRACE [read_logger] Read 65181360-65181363 (4 bytes). Total requests: 4585 (7913636 bytes),read,65181360,65181363,4,4,4585,7913636
TRACE [read_logger] Read 65181364-65181831 (468 bytes). Total requests: 4586 (7914104 bytes),read,65181364,65181831,468,468,4586,7914104
TRACE [read_logger] Read 65181832-65181835 (4 bytes). Total requests: 4587 (7914108 bytes),read,65181832,65181835,4,4,4587,7914108
TRACE [read_logger] Read 65181836-65182287 (452 bytes). Total requests: 4588 (7914560 bytes),read,65181836,65182287,452,452,4588,7914560
```

At I glance, I saw "Total requests: 4588" which would be an absurdly high number of http requests. That's what gave me pause. 

But upon closer inspection, I realized that as well as http-requests, the http-client logs reads from it's in own in memory buffer. There are two slightly different kinds of log lines, the lines are labeled in their second comma delimited field with `read` or `http-range`. `read` corresponds to the in memory buffer, and the `http-read` ones correspond to the number of http requests.

So actually, this bbox query requires 44 http requests, which is pretty reasonable given the dataset / query.

src/rust/Cargo.toml
```diff
@@ -25,6 +25,7 @@ log = "0.4.20"
 fallible-streaming-iterator = "0.1.9"
 tempfile = "3.8.1"
 reqwest = { version = "0.11.22", optional = true }
+async-trait = "0.1.81"
```
bjornharrtell, line 28: why is this added/required?

bjornharrtell (commented):
LGTM just one question

src/rust/Cargo.toml
michaelkirk: Whoops, missed your response last week.

I need `async-trait` for the new test client: https://github.com/flatgeobuf/flatgeobuf/pull/376/files#diff-b1800ddf53707433e40d9040a9eee56ce39144eaeafca4ca60a5a30893fdcbdbR47

Note that it's ultimately not a *new* dependency, it's just explicit now. It's required by http-range-client (which is why it's required for the test client):

```
$ cargo tree
...
├── http-range-client v0.7.2        
│   ├── async-trait v0.1.81 (proc-macro)
```

Since it's only used in the tests, I've moved it to the "dev" dependencies, so thank you for asking about it.  😉

dabreegster:
Just confirming this fix works excellently in for a case with 1000s of tiny polygons packed together. An example goes from >1000 requests (mostly with ~300 byte responses) and over 30s of waiting to ~30 requests and subsecond. Thanks again!

bjornharrtell:
Thanks for testing @dabreegster, and with I will approve and merge. :)

bjornharrtell (approved):

bjornharrtell:
@michaelkirk thanks for this nice improvement!

PR #378, opened 2024-08-10
[rust] update deps

### LINKED ISSUE #367: [rust] Can't upgrade to reqwest 0.12.x
Getting " ^^^^^ the trait `AsyncHttpRangeClient` is not implemented for `Client`"

michaelkirk:
Fixes #367 

~~Draft because it depends on https://github.com/pka/http-range-client/pull/7~~ **edit: merged**

### DIFF
--- a/src/rust/Cargo.toml
+++ b/src/rust/Cargo.toml
@@ -16,15 +16,18 @@ default = ["http"]
 http = ["http-range-client", "bytes", "reqwest"]
 
 [dependencies]
-flatbuffers = "23.5.26"
+# chore: FlatBuffers does not follow SemVer, but rather uses a format of the date of the release.
+# So we should match versions exactly and bump only explicitly.
+# Updates are typically synchronized across all FGB clients, and codegen should be rerun after updating.
+flatbuffers = "=23.5.26"
 byteorder = "1.5.0"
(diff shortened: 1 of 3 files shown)


michaelkirk:
Ok, this is now ready for review.

src/rust/Cargo.toml
```diff
@@ -16,15 +16,15 @@ default = ["http"]
 http = ["http-range-client", "bytes", "reqwest"]
 
 [dependencies]
-flatbuffers = "23.5.26"
+flatbuffers = "24.3.25"
```
bjornharrtell, line 19: I don't think we should bump flatbuffers in this update, that should be done with new code gen and I prefer to do that for all languages with a single PR.

bjornharrtell (changes requested):

src/rust/Cargo.toml
michaelkirk: Good point. 

I didn't realize until now that flatbuffers versioning were simply dates, and irrespective of any semantic versioning, so I made our constraint specific to avoid any potential over-eager patch updates that might not be semver compatible.

bjornharrtell (approved):

bjornharrtell:
Thanks @michaelkirk. :)

PR #386, opened 2024-09-21
[js] add example serialize.js

3-yuma-3:
Thank you for deesrialize.js, it's very helpful!!

Corresponding serialize.js will help developers more.

How about this change?


bjornharrtell:
Might be a helpful example but I think I need to make it more clear that the current ts/js impl. doesn't make spatial index

3-yuma-3:
How can I more clear it?
Should I add some comment in serialize.js to descrive it?

bjornharrtell:
@3-yuma-3 I've added a comment, will merge. Thanks. :)

bjornharrtell (approved):

3-yuma-3:
@bjornharrtell   

Thanks a lot!!

PR #387, opened 2024-09-26
[rust] Upgrade to geozero 0.14

### LINKED ISSUE #283: Resolve circular dependency of Rust implemention with geozero

pka:
The Rust implementation should have a new release, because of major updates in `geozero`. Fixes #283 !

### DIFF
--- a/src/rust/CHANGELOG.md
+++ b/src/rust/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [4.4.0] - UNRELEASED
+
+- [rust] Upgrade to geozero 0.14
+
 ## [4.3.0] - 2024-08-16
 
 - [all] Upgrade to flatbuffers 24.3.25 (#380)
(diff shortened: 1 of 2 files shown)


bjornharrtell:
Published as 4.4.0 - https://crates.io/crates/flatgeobuf/4.4.0

PR #392, opened 2024-10-29
Rust: Add `geo-traits` reader implementation

kylebarron:
We have a new core `geo` crate: [`geo-traits`](https://docs.rs/geo-traits/latest/geo_traits/)! This enables zero-copy vector data interchange throughout the GeoRust ecosystem.

This PR implements these traits on new-types around a `flatgeobuf::Geometry`.

Since Flatbuffers are zero-copy, this means that you could memory-map a FlatGeobuf file and pass those geometry references directly to any consumer accepting these traits.

### Change list 

- Adds a public method onto `FgbFeature` to return a struct that implements `geo_traits::GeometryTrait`.
- Since the method returns an `impl geo_traits::GeometryTrait`, none of the new structs need to be exposed publicly.
- This provides a standard way to access FlatGeobuf data with Z and M values as well.

### Questions

- How to handle FlatGeobuf files with `t` values? Those are currently omitted
- How should we handle curve types? This PR currently panics on those.

### DIFF
--- a/src/rust/src/error.rs
+++ b/src/rust/src/error.rs
@@ -10,6 +10,7 @@ pub enum Error {
     IllegalHeaderSize(usize),
     InvalidFlatbuffer(InvalidFlatbuffer),
     IO(std::io::Error),
+    UnsupportedGeometryType(String),
 }
 pub type Result<T> = std::result::Result<T, Error>;
 
@@ -23,6 +24,7 @@ impl Display for Error {
             Error::IllegalHeaderSize(size) => write!(f, "Illegal header size: {size}"),
             Error::InvalidFlatbuffer(invalid_flatbuffer) => invalid_flatbuffer.fmt(f),
             Error::IO(io) => io.fmt(f),
+            Error::UnsupportedGeometryType(s) => f.write_str(s),
         }
     }
 }
--- a/src/rust/src/properties_reader.rs
+++ b/src/rust/src/properties_reader.rs
@@ -42,7 +42,9 @@ impl FgbFeature {
     ///
     /// Note that this method only supports the core geometry types supported by [geo_traits], and
     /// will panic on curve geometries.
-    pub fn geometry_trait_impl(&self) -> Result<Option<impl geo_traits::GeometryTrait + use<'_>>> {
+    pub fn geometry_trait_impl(
+        &self,
+    ) -> Result<Option<impl geo_traits::GeometryTrait<T = f64> + use<'_>>> {
         if let Some(geom) = self.geometry() {
             let dim = self.dimension();
             let result = match self.header().geometry_type() {
--- a/src/rust/src/properties_reader.rs
+++ b/src/rust/src/properties_reader.rs
@@ -42,37 +42,45 @@ impl FgbFeature {
     ///
     /// Note that this method only supports the core geometry types supported by [geo_traits], and
     /// will panic on curve geometries.
-    pub fn geometry_trait_impl(&self) -> Option<impl geo_traits::GeometryTrait + use<'_>> {
-        let geom = self.geometry()?;
-        let dim = self.dimension();
-        let result = match self.header().geometry_type() {
-            GeometryType::Point => {
-                crate::geo_trait_impl::Geometry::Point(crate::geo_trait_impl::Point::new(geom, dim))
-            }
-            GeometryType::LineString => crate::geo_trait_impl::Geometry::LineString(
-                crate::geo_trait_impl::LineString::new(geom, dim),
-            ),
-            GeometryType::Polygon => crate::geo_trait_impl::Geometry::Polygon(
-                crate::geo_trait_impl::Polygon::new(geom, dim),
-            ),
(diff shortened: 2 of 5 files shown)


src/rust/src/properties_reader.rs
```diff
@@ -27,6 +27,53 @@ impl FgbFeature {
...
+            GeometryType::MultiPolygon => crate::geo_trait_impl::Geometry::MultiPolygon(
+                crate::geo_trait_impl::MultiPolygon::new(geom, dim),
+            ),
+            GeometryType::Unknown => crate::geo_trait_impl::Geometry::new(geom, dim),
+            GeometryType::GeometryCollection => {
+                crate::geo_trait_impl::Geometry::GeometryCollection(
+                    crate::geo_trait_impl::GeometryCollection::new(geom, dim),
+                )
+            }
+            _ => todo!("How should we handle other geometry types here?"),
```
kylebarron, line 73: I think this is the primary remaining question in this PR before it can receive a full review: should this function return a result? Should this error on non-simple geometry types?

pka: I think returning a `Result` is better, since you could also have a non-supported geometry type in a GeometryCollection.

src/rust/src/properties_reader.rs
```diff
@@ -27,6 +27,61 @@ impl FgbFeature {
...
+            (false, false) => geo_traits::Dimensions::Xy,
+        }
+    }
+
+    /// Access the underlying geometry, returning an object that implements
+    /// [geo_traits::GeometryTrait].
+    ///
+    /// Note that this method only supports the core geometry types supported by [geo_traits], and
+    /// will panic on curve geometries.
+    pub fn geometry_trait_impl(&self) -> Result<Option<impl geo_traits::GeometryTrait + use<'_>>> {
```
kylebarron, line 45: I learned in https://github.com/kylebarron/wkb that this `use<'_>` syntax requires Rust 1.82. https://rust-lang.github.io/rfcs/3617-precise-capturing.html

This signature would be nice to use as it lets all of `geo_trait_impl` be a private module. I don't see a current MSRV:
https://github.com/kylebarron/flatgeobuf/blob/7e0f27e50ddde197164afcb532132796c775609a/src/rust/Cargo.toml#L2-L13

Thoughts?

src/rust/src/properties_reader.rs
```diff
@@ -27,6 +27,61 @@ impl FgbFeature {
...
+                    crate::geo_trait_impl::MultiPolygon::new(geom, dim),
+                ),
+                GeometryType::Unknown => crate::geo_trait_impl::Geometry::new(geom, dim),
+                GeometryType::GeometryCollection => {
+                    crate::geo_trait_impl::Geometry::GeometryCollection(
+                        crate::geo_trait_impl::GeometryCollection::new(geom, dim),
+                    )
+                }
+                geom_type => {
+                    return Err(GeozeroError::Geometry(format!(
```
kylebarron, line 74: Should this return `flatgeobuf::Error` instead of `GeozeroError`? I'd think it should, however none of the `flatgeobuf::Error` variants are obvious to use:
```
pub enum Error {
    MissingMagicBytes,
    NoIndex,
    HttpClient([HttpError](https://docs.rs/http-range-client/0.8.0/x86_64-unknown-linux-gnu/http_range_client/error/enum.HttpError.html)),
    IllegalHeaderSize([usize](https://doc.rust-lang.org/nightly/std/primitive.usize.html)),
    InvalidFlatbuffer([InvalidFlatbuffer](https://docs.rs/flatbuffers/24.3.25/x86_64-unknown-linux-gnu/flatbuffers/verifier/enum.InvalidFlatbuffer.html)),
    IO([Error](https://doc.rust-lang.org/nightly/std/io/error/struct.Error.html)),
}
```

src/rust/src/properties_reader.rs
pka: Adding an MSRV 1.82 is fine for me.

src/rust/src/properties_reader.rs
pka: Yes, we should add an FGB Error for that.

kylebarron:
Note that we may want to wait for geo-traits 0.2 before merging and releasing: https://github.com/georust/geo/pull/1263

pka (approved):
A usage example would be nice.

bjornharrtell:
@kylebarron seems like very nice work (even though I seem to get less rather than more into rust, unfortunately) - I'll merge now and see if I can make it into a published crate not too far off.

src/rust/src/error.rs
```diff
@@ -10,6 +10,7 @@ pub enum Error {
     IllegalHeaderSize(usize),
     InvalidFlatbuffer(InvalidFlatbuffer),
     IO(std::io::Error),
+    UnsupportedGeometryType(String),
```
michaelkirk, line 13: Note that adding a new error variant is a breaking change unless the enum is marked as non-exhaustive.

PR #393, opened 2024-11-04
[ts] fix missing binary typed properties

Luavis:
The FGB specification and the column type declarations included binary properties, but these properties were not implemented in the TypeScript code for de/serialization.

This commit adds the necessary logic to handle parsing of binary typed properties, ensuring full compliance with the FGB specification.


bjornharrtell (approved):
Thanks @Luavis, it looks good

bjornharrtell (approved):

PR #404, opened 2024-12-16
[rust] Allow downstream crates to choose TLS provider

kylebarron:
Allow downstream crates to choose their own TLS provider.

Depends on https://github.com/pka/http-range-client/pull/9.

Currently, all downstream crates are forced into reqwest's `default-tls` because that's what `http-range-client` uses. This forces the use of openssl, which makes it harder to package for Python. I'd like to use [rustls](https://github.com/rustls/rustls) instead for a pure-Rust solution.


bjornharrtell:
LGTM, @pka and @michaelkirk are you ok with it?

michaelkirk:
Do you have an example where you're using rustls? 

I was expecting you'd need to plumb additional feature flags for each potential tod backend. Am I mistaken?

kylebarron:
> I was expecting you'd need to plumb additional feature flags for each potential tls backend. Am I mistaken?

No, I don't think so, because cargo will automatically merge feature flags. So if I separately depend on `reqwest` with a rustls feature, I think the unification will just work. In my case I'm also using `object_store`, which turns on the rustls feature flags itself: https://github.com/apache/arrow-rs/blob/c4dbf0d8af6ca5a19b8b2ea777da3c276807fc5e/object_store/Cargo.toml#L53

With the current flags, it requires downstream users to have openssl available, see https://github.com/geoarrow/geoarrow-rs/pull/950.

michaelkirk:
You're right!  I was overly conservative on how resolver=2 works.

I set up an example here to convince myself: http://github.com/michaelkirk/gh_fgb_404_feature_unification_example

Once https://github.com/pka/http-range-client/pull/9 is released, we can bump our min requirement here and merge this.

kylebarron:
Temporarily using the latest main of http-range-client on this branch so I can test from geoarrow-rs.

kylebarron:
I verified in https://github.com/geoarrow/geoarrow-rs/pull/950 that this change enables compilation in environments that don't have access to OpenSSL. 

I think this also enables using the async Flatgeobuf reader in WebAssembly but haven't tested that yet.

michaelkirk:
Some years ago we integrated FGB into https://github.com/a-b-street/abstreet/blob/main/popdat/Cargo.toml - I think @dabreegster might have used it elsewhere too.

IIRC http, and (thus tls), is the responsibility of the transport layer - in this case reqwest, which ultimately calls out to the browser on wasm (again, IIRC).

It might be useful to some people to support non-browser / non-reqwest targets though.

kylebarron:
I don't have an exact repro now, but I failed previously at getting async FlatGeobuf working in WebAssembly; I had only gotten sync parsing from a buffer.

> IIRC http, and (thus tls), is the responsibility of the transport layer - in this case reqwest, which ultimately calls out to the browser on wasm (again, IIRC).
> 
> It might be useful to some people to support non-browser / non-reqwest wasm targets though.

To be clear, I don't use `reqwest` directly. I [implement `AsyncHttpRangeClient` for `ObjectStore`](https://github.com/geoarrow/geoarrow-rs/blob/d61e866e6cef54f6538c3f0841bc94c479e40dfb/rust/geoarrow/src/io/flatgeobuf/reader/object_store_reader.rs), which lets me use [`ObjectStore`](https://docs.rs/object_store/latest/object_store/) to transparently read FlatGeobuf files on S3/GCS/Azure, etc. In normal server use, `object_store` happens to ultimately use `reqwest` under the hood to send the raw HTTP calls. But for other `object_store` reasons, `object_store` with the `http` feature doesn't compile for wasm today, so we have a [separate implementation](https://github.com/H-Plus-Time/object-store-wasm) of the `ObjectStore` trait that works for S3 and HTTP files from Wasm.

This is nice because geoarrow only needs to use a single library's APIs for fetching remote files, whether it's in Python, Rust, or in Wasm.

kylebarron:
http-range-client 0.9 has now been released, so this PR is unblocked.

pka (approved):

PR #406, opened 2024-12-19
add success and failure callbacks to loader

riley-nova:
Issue reported here: https://github.com/flatgeobuf/flatgeobuf/discussions/396

openlayers loader requires a success or failure callback to fire the 'featuresloadend' and 'featuresloaderror' events. The current loader implementation does not use them. This is documented in the VectorSource api: https://openlayers.org/en/latest/apidoc/module-ol_source_Vector-VectorSource.html

> "The loader function used to load features, from a remote source for example. If this is not set and url is set, the source will create and use an XHR feature loader. The 'featuresloadend' and 'featuresloaderror' events will only fire if the success and failure callbacks are used."

This might not be the most elegant solution (wrapping the loader in a try/catch) but I didnt want to dig to deep into all the deserialize functions and add something there.


bjornharrtell (changes requested):
I think it's fine with a try catch here and use of the callbacks. @riley-nova can you fix the format and type issues?

bjornharrtell (approved):

bjornharrtell:
Thanks @riley-nova, it looks good to me, will merge now.

PR #410, opened 2025-01-10
[all] Upgrade to flatbuffers 24.12.23

bjornharrtell:


michaelkirk:
Want me to take a look at the rust failures @bjornharrtell?

(No problem if you want to do it yourself though 😉)

bjornharrtell:
Feel very free @michaelkirk 🙂

michaelkirk:
I ran out of time today, but I can take a look next week if you don’t get to it first.

michaelkirk:
Sorry that I was too slow, and thanks for addressing this. 

The namespace feature seems weird when applied to the rust codegen in particular when you have multiple .fbs files - I agree that commenting it out seems reasonable for rust.

I'm a little confused as to why flatc started outputting the namespace as a module into the rust code only now, but whatever.

bjornharrtell:
@michaelkirk I manually handled it in the past, but forgot. Now I have scripted it.

PR #411, opened 2025-02-07
[java] Make JTS dependency allow up to 2.0

nicolas-f:
Java code, upgrade JTS dependency version to 1.20

### DIFF
--- a/src/java/pom.xml
+++ b/src/java/pom.xml
@@ -43,7 +43,7 @@
     <dependency>
       <groupId>org.locationtech.jts</groupId>
       <artifactId>jts-core</artifactId>
-      <version>1.20.0</version>
+      <version>[1.19,2.0)</version>
     </dependency>
     <dependency>
       <groupId>junit</groupId>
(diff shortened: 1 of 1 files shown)


bjornharrtell:
@nicolas-f I see.. is there no better way to handle this, i.e in a way that will allow the dependency on jts to be 1.19 or above  so that it will also work when jts releases 1.21?

nicolas-f:
> @nicolas-f I see.. is there no better way to handle this, i.e in a way that will allow the dependency on jts to be 1.19 or above so that it will also work when jts releases 1.21?

Ah yes I see, it is possible, I never use it actually:

You want something like that ? :

https://books.sonatype.com/mvnref-book/reference/pom-relationships-sect-project-dependencies.html#pom-relationships-sect-version-ranges

bjornharrtell:
@nicolas-f yeah it seems like a good idea to me because it's very unlikely JTS will make a breaking change within the major version.

bjornharrtell:
@nicolas-f LGTM, thanks! :)

bjornharrtell (approved):

PR #414, opened 2025-03-11
Update README.md

petersmythe:
Helps readers quickly figure out which is better


bjornharrtell:
thanks @petersmythe :)

PR #418, opened 2025-03-20
TS: Support for XYZ, XYM & XYZM layouts

bkuster:
# Issue

Reading and writing from openlayers only supports 2D coordinates, even though both openlayers and flatgeobuf support other geometry layouts. 

# Solution

This MR add the capabilities to read & write all openlayer layouts using the node library.

### DIFF
--- a/src/ts/generic/geometry.ts
+++ b/src/ts/generic/geometry.ts
@@ -117,7 +117,6 @@ export function parseGeometry(geometry: ISimpleGeometry, headerGeomType: Geometr
         type = toGeometryType(geometry.getType());
     }
 
-
     let flatEnds: number[] | undefined;
     if (type === GeometryType.MultiLineString) {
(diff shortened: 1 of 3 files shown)


bkuster:
i replaced the bitwise *2 multiplier with a * (since XYZ / XYM require 3) but _could_ re-implement for 2 & 4 if you see a large performance benefit.

bjornharrtell (approved):
Looks clean and can perhaps be accepted as is.

The reason I didn't implement it originally is that I find that I almost never want to use other than x/y in a web map. But, it does make sense to support.

I don't see immediate need to optimize further.

bjornharrtell:
@bkuster nice code but there are some lint issues. :)

bkuster:
ah, yes, the linter, will get right onto that. [we](https://github.com/virtualcitySYSTEMS/) develop [3D web maps](https://github.com/virtualcitySYSTEMS/map-core) and have just added flatgeobuf support, hence the need for proper 3D handling.

Issue #416, opened 2025-03-17
[ts] Missing spatial filtering capability when consuming from local file or stream instead of url


bjornharrtell:

bviebahn:
I'm also running into this issue. I have a local fgb file that takes more than 5 seconds to fully deserialize using the geojson deserialize function. Would be great if the spatial filter would work, any chance this could be added?

bjornharrtell:
@bviebahn my personal priority to do this is currently very low, because I prefer other languages when running things locally. But feel free to contribute in any way.

PR #421, opened 2025-03-31
[ts] Add support for spatial filtering on Uint8Array

bviebahn:
Related issue #416 

This adds support for spatial filtering when deserializing from a Uint8Array. There are probably a few things that could be cleaned up and simplified. I'm not very familiar with the codebase, so I tried to stick close to the existing implementation that's using the `HttpReader` and `streamSearch` functions.

I created a `ArrayReader` class that's very similar to the `HttpReader`, except that the `selectBbox` method is not an async generator function. I also created another `search` function that's basically an exact copy of `streamSearch` except it's also not an async generator.

Any suggestions how to clean this up or simplify this are appreciated. (Or feel free to make any changes yourself)


bjornharrtell:
@bviebahn looks sensible, execept that I would think it could be made async in some way so that streamSearch can still be used. Or, do you have any arguments for why it needs the duplicated logic in search?

bviebahn:
@bjornharrtell Yes, we could also just make it async, I just didn't want to change the signature of the deserialize functions, since that would be a breaking change. But if it's ok with you I can make that change. So then all deserialize functions would just return async generators.

bjornharrtell:
@bviebahn ah yes... yeah I think we should drop the IFeature[] return support and just go with async, even if that means a breaking change. Would of course be good to get deserializeStream supporting filtering too so "NOTE: Does not support spatial filtering" can be removed.

bviebahn:
Sure, I'll look into that today!

bjornharrtell:
@bviebahn sounds great, thanks for your efforts!

bviebahn:
@bjornharrtell made the changes, now all deserialze implementation returns async generators.
Regarding adding support for spatial filtering to deserializeStream, it doesn't seem to be as straightforward, because R-tree traversal requires random access, right? As I see it the only way around that would be to read the entire stream first, then we could use the same ArrayReader for filtering, but that kind of defeats the purpose of using a stream, right?

bjornharrtell:
@bviebahn no the spatial index search can be done with forward seek only. :) But I can see it's not a trivial effort to add to this PR, so probably better to do that later if possible.

bviebahn:
Ah ok, I didn't know that. Yes, I would prefer to leave it for another time then if that's ok. Anything else?

bjornharrtell (approved):
LGTM

bviebahn:
@bjornharrtell Nice, thanks! Could we update the npm package?

bjornharrtell:
@bviebahn yeah will get it done hopefully sooner than later.

bjornharrtell:
@bviebahn https://www.npmjs.com/package/flatgeobuf/v/4.0.0 is now available.

Issue #436, opened 2025-07-08
Feature Request: Bearer Authorization support (with range requests)


danielsparing:
While deserialize is [able](https://github.com/flatgeobuf/flatgeobuf/blob/master/examples/maplibre/large.html#L83) to send range requests to a URL, it would be even nicer if it was possible to include a bearer token to authorize with.

Or is this somehow already possible?

bjornharrtell:
Done and published as 4.1.1

PR #437, opened 2025-07-10
[ts] Allow custom fetch headers through high level API

bjornharrtell:
ref https://github.com/flatgeobuf/flatgeobuf/issues/436

### DIFF
--- a/src/ts/generic.ts
+++ b/src/ts/generic.ts
@@ -43,10 +43,11 @@ export function deserialize(
     fromFeature: FromFeatureFn,
     rect?: Rect,
     nocache = false,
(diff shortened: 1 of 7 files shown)


bjornharrtell:
This might do the trick @danielsparing, what do you think?

danielsparing:
Thank you so much for such a quick turnaround! As far as I can tell, looks promising, I'm not smart enough to test from within a branch, but would love to try once makes it to release. fyi @jkoelewijn-dat

danielsparing:
@bjornharrtell I finally got around to testing it and it works great, thanks!

PR #440, opened 2025-07-22
Rust: Prepare 5.0 release

kylebarron:
Including #439 

It would be nice to have a new release with geo-traits 0.3 so that I can release the [`geoarrow-flatgeobuf` crate](https://github.com/geoarrow/geoarrow-rs/tree/main/rust/geoarrow-flatgeobuf), which loads FlatGeobuf local or remote files into GeoArrow in memory.

Bumping to geo-traits 0.3 is technically a breaking change, but `flatgeobuf` 4.4.0 bumped to `geozero` 0.14, so it doesn't seem like `flatgeobuf` exactly follows semver? cc @michaelkirk


bjornharrtell:
I think the intent was to follow semver. I do that in other implementations.

kylebarron:
Ok, I think that means that whenever any dependency has a breaking release, like `geo-traits` bumps to `0.3` or `geozero` bumps to 0.14, then I think you'd need a new major version of `flatgeobuf`?

kylebarron:
Ok, in the context of this discussion I changed this PR to update the crate version to 5.0 instead of 4.7.

kylebarron:
Or... is there a way to do this non-breaking by having feature flags to choose which version of `geo-traits` to link?

kylebarron:
Is there some prep work to do in a 5.0 release to ensure that we could bump to a future geo-traits 0.4 version while not needing a breaking release? I.e. should we put `geo-traits` behind a feature flag that is turned on by default?

kylebarron:
The geo-traits 0.2 to 0.3 migration should be relatively simple, and there are probably relatively few users of `flatgeobuf` using the `geo-traits` integration thus far. So it should be an easy upgrade for most `flatgeobuf` crate users.

bjornharrtell (approved):

kylebarron:
Do you have plans to publish 5.0 to crates.io?

kylebarron:
Thank you!

PR #442, opened 2025-08-05
POC: Mutable Flatgeobuf

logan-keede:


logan-keede:
Sorry meant to make this on a local fork, but it should be fine as POC PR

PR #444, opened 2025-08-20
Allow AsyncFeatureEnumerator to read files w/o a CRS definition or no geometry

mlptownsend:
If you create a flatgeobuf file with no CRS definition or with no geometry at all, AsyncFeatureEnumerator.Create will throw exceptions and fail. This PR adds some null checks to allow the files to be opened.

Interestingly, when you have a file with no geometry, there's no header information to tell you how many records are in the file. You just have to iterate and count in this situation.

To generate some sample files with these problems, we can have DuckDB do it w/ its spatial extension.
```
INSTALL spatial;
LOAD spatial;

COPY(
select * from st_read('path/to/test/data/countries.fgb')
) to 'path/to/test/data/countries_nocrs.fgb' (FORMAT GDAL, DRIVER 'FlatGeoBuf')

COPY(
select * EXCLUDE(geom) from st_read('path/to/test/data/countries.fgb')
) to 'path/to/test/data/countries_nogeo.fgb' (FORMAT GDAL, DRIVER 'FlatGeoBuf', LAYER_CREATION_OPTIONS 'SPATIAL_INDEX=NO')
```


bjornharrtell:
Thanks @mlptownsend seems like a solid fix.

mlptownsend:
I just realized that without this fix, you can't reopen a file created with FeatureCollectionConversions.SerializeAsync because there's no way to set the header's CRS with this method.

```
[TestMethod]
public async Task TestRoundtripAsyncToSyncToAsync() {
    var ae = await AsyncFeatureEnumerator.Create(File.OpenRead("../../../../../../test/data/countries.fgb"));
    var allFeatures = new System.Collections.Generic.List<NetTopologySuite.Features.IFeature>();
    while (await ae.MoveNextAsync()) {
        allFeatures.Add(ae.Current);
    }

    using var msOut = new System.IO.MemoryStream();
    await FeatureCollectionConversions.SerializeAsync(msOut, allFeatures, GeometryType.MultiPolygon);
    msOut.Position = 0;

    //Fails w/ Crs null ref
    var ae2 = await AsyncFeatureEnumerator.Create(msOut);
}
```

PR #445, opened 2025-08-20
Add method to AsyncFeatureEnumerator to skip a number of records without allocating every feature along the way.

mlptownsend:
Added SkipAsync() to AsyncFeatureEnumerator to skip records without incurring the allocation and time cost of creating each feature it encounters on the way there. I have a process that can be resumed and it'd be nice to be able to jump back in faster. Stream.ReadAsync still needs a byte buffer to advance so we still need the array pool buffer, can't avoid that.


bjornharrtell:
Thanks @mlptownsend this seems like a sensible addition.

PR #447, opened 2025-08-22
Support serializing and deserializing binary attribute values.

mlptownsend:
Added support to the C# library for binary attributes. Currently it will throw an exception if you try and read a file with a byte array in the feature's attribute table. It also ties a bit into the earlier PR I made for the no geometry case. I also created a test file that contains an "id" string field and a "wkb" byte array field. 

I had to add a case to FeatureCollectionConversions.ToColumnType to check for a byte array instead of the typecode (which returns an Object). In GeometryConversions.FromFlatbuf, I added a check to return null if the provided FlatBuf geometry was unknown and all its lengths are zero, reasoning that it probably means it's completely blank. Then in FeatureConversions, if the header geometry is unknown and the feature geometry is null, then it's going to be a null geometry. And then I figured out that the binary's byte array value works exactly the same as it does for the strings. I changed the UTF8.GetString call in FromFeature to use the original propertiesArray to avoid a memory copy w/ ToArray'ing the whole memory stream.


src/net/FlatGeobuf/NTS/FeatureConversions.cs
```diff
@@ -16,8 +16,10 @@ public static ByteBuffer ToByteBuffer(IFeature feature, HeaderT header)
             GeometryType geometryType;
             if (header.GeometryType != GeometryType.Unknown)
                 geometryType = header.GeometryType;
-            else
+            else if (feature.Geometry is not null)
```
bjornharrtell, line 19: This seem unrelated to the intent of this PR?

src/net/FlatGeobuf/NTS/GeometryConversions.cs
```diff
@@ -256,6 +256,8 @@ public static NTSGeometry FromFlatbuf(GeometryFactory factory, FlatGeobufCoordin
 
             if (type == GeometryType.Unknown)
                 type = geometry.Type;
+            if (type == GeometryType.Unknown && geometry.XyLength + geometry.ZLength + geometry.MLength + geometry.TLength + geometry.TmLength + geometry.PartsLength + geometry.EndsLength == 0)
```
bjornharrtell, line 259: This seem unrelated to the intent of this PR?

src/net/FlatGeobuf/NTS/FeatureConversions.cs
bjornharrtell: I see there is a tie in with the empty/null geometry

src/net/FlatGeobuf/NTS/GeometryConversions.cs
bjornharrtell: I see there is a tie in with the empty/null geometry

Issue #448, opened 2025-08-26
[ts] Publish generic parseProperties function


ger-benjamin:
Hi,

I worked a bit with this FlatGeoBuf and Openlayers and I was able to read my `.fgb` files, and display my features. This format is really performant and nice to use.

But I also wanted to write custom features and that was a bit tedious.
I've started with the ol (openlayers) part.

The `fromFeature` function is defined in the `src/ol` repository, not exported outside the library, not configurable and passed directly to the `genericFromFeature` and `deserialize` function:

- https://github.com/flatgeobuf/flatgeobuf/blob/master/src/ts/ol/feature.ts#L4
- https://github.com/flatgeobuf/flatgeobuf/blob/master/src/ts/ol/featurecollection.ts#L10
- https://github.com/flatgeobuf/flatgeobuf/blob/master/src/ts/ol.ts#L40

The very good point is that it's still possible to re-implement everything without a lot of code using what is exported by the the `src/generic.ts` file.

The only method I had to copy-paste from this repo to my repo was the parseProperties method: https://github.com/flatgeobuf/flatgeobuf/blob/master/src/ts/generic/feature.ts#L159

Without copy-pasting this method, I can't write my own `fromFeature` function to pass to the `generic.deserialize` function. (Or I can have custom features, but without properties).

It would be great to publish this method to ease the creation of custom features from the generic Features.

I'll propose a PR.

PR #449, opened 2025-08-26
[ts] Publish generic parseProperties function

ger-benjamin:
For issue #448 

Publish `generic/feature.ts parseProperties` function to ease the creation of custom features from generic Features.


ger-benjamin:
@bjornharrtell that should be fixed, sorry, I didn't see this target before.

Issue #455, opened 2025-09-04
Add possibility to create ol RenderFeature


ger-benjamin:
**Context**

Currently, the `ol` part can deserialize FlatGeoBuf file into ol/Feature. This work well, but ol/Features are heavy and need a lot of RAM.

Since some versions, ol introduced `ol/render/Feature` (RenderFeature), that are **light readonly Features**. This work well typically with WebGL layers that allows to display several ten-thousands of feature on a standard computer (and million(s) on a good computer, which would be impossible with standard ol/feature).

FlatGeoBuf works well with big datasets, it would be great to take benefit of RenderFeatures.

**The easy part**

Creating RenderFeature with Openlayers is quite easy. That's just another constructor to call on demand.

**Difficulties**

One difficulty is that RenderFeature are read-only. Every properties must be set at the feature creation.

Combined to that, the flatgeobuf/ol/feature.fromFeature function (https://github.com/flatgeobuf/flatgeobuf/blob/4019ebd78f15d356d8cdca0a533be4dc5ec93714/src/ts/ol/feature.ts#L18) is used by the feature-collection and directly passed to the generic part to create the features. It's a bit hard to enter in this chain of functions to add a choice of creating Feature or RenderFeature.

Also, there is already a lot of parameter on the entry-point function https://github.com/flatgeobuf/flatgeobuf/blob/4019ebd78f15d356d8cdca0a533be4dc5ec93714/src/ts/ol.ts#L79-L85

**Implementation proposition 1**

One possibility is to add a "FeatureClass" parameters and to transmit this parameters in all chained ol functions.
And then, adapt the flatgeobuf/ol/feature.fromFeature function to render a ol/Feature or a RenderFeature.

But that's a lot of repetition. We would have to add a parameters `featureClass` in a lot of function...

**Implementation proposition 2**

To limit the number of parameters to pass from functions to functions, and to work a bit more like the ol formats are working (https://openlayers.org/en/latest/apidoc/module-ol_format_GeoJSON-GeoJSON.html), I propose to export a `FeatureCollection` class instead of exporting independently the serialize and deserializes functions from the ol/FeatureCollection format.

This class can have some options (FeatureClass, dataProjection, featureProjection, nocache, headers....) that are currently passed from function to function.

Based on the `FeatureClass` option, the deserialize methods could create Feature or RenderFeature.

[...]

PR #457, opened 2025-09-09
Add OL RenderFeature creation possibility

ger-benjamin:
WIP

For #455 - Add ol RenderFeature support

- [x] Adapt Code
- [x] Adapt vite examples
- [ ] Adapt tests
- [ ] Review

Now the `flatGeobuff/src/ol/featurecollection.ts` export a class with `serialzize`, `deserialize` and `getRect` (extent/bbox) methods.

This class comes with options to lower the number of arguments needed in every functions.

One option is `featureClass`. This allows to ask the class to deserialize fgb files into `Feature`s (default) or `RenderFeature`s.

The `flatGeobuff/src/ol/feature.ts` file now have functions to create `Feature`s or `RenderFeature`s.

The  options are inspired by ol/format (like the [GeoJSON](https://openlayers.org/en/latest/apidoc/module-ol_format_GeoJSON-GeoJSON.html) format that also has a `featureClass` option to display `RenderFeatures`, dataProjection and featureProjection)


bjornharrtell:
@ger-benjamin yes I didn't mean I want to remove forever or that I don't think it's a worthwile addition. :) So perhaps a rollback and  4.x release and reintroduce as 5.x as you suggest. But I will look into it a bit first to understand the change better. No panic.

PR #461, opened 2025-11-21
Java Z and M dimensions serialization

### LINKED ISSUE #460: [java] Use proper Coordinate class in GeometryConversions
Hi everybody,
I was investigating a case where a flatgeobuf file exported from java was way bigger than the equivalent produced using gdal's `ogr2ogr`.

I found that `GeometryConversions#deserialize` is always using the `CoordinateXYZM` or `Coordinate` classes; in my case I only have `X,Y` and a `Coordinate` is used; this implies that the geometry is deemed to have 3 dimensions in `GeometryConversions.ReverseXYCoordinateSequenceFilter#filter` because `seq.hasZ()` returns true, even if the z value is set to `Coordinate.NULL_ORDINATE`.

This makes the generated files to grow from 650megs to 820megs.

I created a custom deserializer, only changing [the deserialize method, at line 234](https://github.com/flatgeobuf/flatgeobuf/blob/87ce48755858431100621b0ed1bf1dcdb6155d2a/src/java/src/main/java/org/wololo/flatgeobuf/GeometryConversions.java#L234)

repacing that block with:
```java
        int c = 0;
        for (int ii = 0, i = 0; ii < xyLength; ii += 2, i++) {
            if (i >= geometry.zLength() && i >= geometry.mLength()) {
                coordinates[c++] = new CoordinateXY(geometry.xy(ii), geometry.xy(ii + 1));
            } else if (i < geometry.zLength() && i < geometry.mLength()) {
                coordinates[c++] = new CoordinateXYZM(geometry.xy(ii), geometry.xy(ii + 1), geometry.z(i),
                        geometry.m(i));
            } else if (i < geometry.zLength()) {
                coordinates[c++] = new Coordinate(geometry.xy(ii), geometry.xy(ii + 1), geometry.z(i));
            } else {
                coordinates[c++] = new CoordinateXYM(geometry.xy(ii), geometry.xy(ii + 1), geometry.m(i));
            }
        }
```

This uses the correct `Coordinate` class, and then the generated file has the expected size.

Happy to open a PR if needed.

tnolli:
fixes #460

### DIFF
--- a/src/java/src/main/java/org/wololo/flatgeobuf/GeometryConversions.java
+++ b/src/java/src/main/java/org/wololo/flatgeobuf/GeometryConversions.java
@@ -24,35 +24,34 @@ import org.wololo.flatgeobuf.generated.GeometryType;
 import com.google.flatbuffers.FlatBufferBuilder;
 
 public class GeometryConversions {
-    private static final GeometryFactory GF = new GeometryFactory();
-
     public static GeometryOffsets serializePart(FlatBufferBuilder builder, org.locationtech.jts.geom.Geometry geometry,
-            int geometryType) throws IOException {
+                                                int geometryType) throws IOException {
         GeometryOffsets go = new GeometryOffsets();
 
         if (geometry == null)
             return go;
 
         switch (geometryType) {
-            case GeometryType.MultiLineString: {
-                int end = 0;
-                MultiLineString mls = (MultiLineString) geometry;
-                if (mls.getNumGeometries() > 1) {
-                    go.ends = new long[mls.getNumGeometries()];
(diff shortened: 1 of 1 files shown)


bjornharrtell (commented):
There is alot of formatting changes here, I have difficulty determining actual semantic difference. While it might be time to do a overhaul in the formatting, I would prefer a PR that changes behaviour only.

tnolli:
No problem, I'll remove the formatting.

Do you also prefer to split the PR into two different ones: one for the serialization problem and one related to the deserialization, to avoid allocating a GeometryFactory for each geometry and, instead, always use the same one (this is thread safe)?

bjornharrtell:
It would be great as separate PRs 🙂

PR #465, opened 2025-11-26
Remove likely false claim of alignment

bjornharrtell:


bjornharrtell:
Note that I think this is rarely a problem anyway AFAIK due to :

1. cross feature direct access is rare
3. in most cases/hardare unaligned access will work anyway

Issue #467, opened 2025-12-01
[ts] Large integers are improperly typed as int32 instead of Long when serializing from GeoJSON FeatureCollection


greenlaw:
When serializing a GeoJSON FeatureCollection to Flatgeobuf, [generic.featurecollection.valueToType](https://github.com/flatgeobuf/flatgeobuf/blob/f0effe7ed85ab22d043bbf86c1d9e7062f75811d/src/ts/generic/featurecollection.ts#L185-L196) is called, which contains this logic:

```
    if (typeof value === 'number') {
        if (value % 1 === 0) return ColumnType.Int;
        return ColumnType.Double;
    }
```

For negative integer values smaller than `-2147483648` and positive values larger than `2147483647`, this results in integer overflow.  `ColumnType.Long` should probably be used in this case instead.

PR #468, opened 2025-12-03
[ts] Encode number as double

bjornharrtell:
ref https://github.com/flatgeobuf/flatgeobuf/issues/467


greenlaw:
Really appreciate you addressing this so quickly @bjornharrtell !  

Are there any plans to cut a minor release soon? We have a production system that would benefit from the fix.

Issue #469, opened 2025-12-04
[ts] fetch only Bounds or Metadata from remote header file


stefanocudini:
Hi,
I was wondering if there's a method in the TypeScript version to retrieve a remote file's extent without downloading all features of a large FGB file.

Looking at the documentation, there's a callback called `headerMetaFn` of `deserialize` method:
```
deserialize(
  url: string,
  rect?: Rect,
  headerMetaFn?: HeaderMetaFn,
```
The problem with this callback is that it seems to be tied to parsing the entire file, I'm starting from a situation where on the client side I don't know the `rect` but only its URL.

From some tests and the code in the examples, this callback returns an `envelope` field that seems to always be null; I assume it wasn't implemented in the ts version.
But aside from this detail, the main problem is that you can't get information about the full extension of the remote file without first parsing it entirely locally.
Yet, analyzing the structure of the binary file, there appears to be a header that also contains some information about the envelope here:
https://github.com/flatgeobuf/flatgeobuf/blob/master/src/fbs/header.fbs#L69

I'm open to suggestions for PRs to submit to get this feature in the TypeScript version.
Any forks that contain this feature, if not yet official, would also be helpful.

Stef

bjornharrtell:
Huh? You don't have to download the full dataset to get the header data. See fx. https://flatgeobuf.org/examples/leaflet/large.html.

stefanocudini:
> Huh? You don't have to download the full dataset to get the header data. See fx. https://flatgeobuf.org/examples/leaflet/large.html.

I noticed that the envelope field is always null. I fixed it in my fork, and now this information is returned.
I'm starting from a use case where I don't know the envelopes or crs of the remote FGB files, so I need an initial query to get only the metadata on which I can then perform partial filtered queries.
In the implementation of your example, you seem to start from a condition where you already know the envelope of your dataset in fact you set the map viewport to cover the dataset extent.

PR #471, opened 2025-12-10
[ts] new method readMetadata()

stefanocudini:
solve issue #469 

Added new method `readMetadata` to get from remote FlatGeobuf resource metadata reading only the file header

- return envelope from metadata previous ignored

- created an example metadata in ViteJs

### DIFF
--- a/src/ts/generic/featurecollection.ts
+++ b/src/ts/generic/featurecollection.ts
@@ -11,12 +11,11 @@ import { Header } from '../flat-geobuf/header.js';
 import type { HeaderMetaFn } from '../generic.js';
 import type { HeaderMeta } from '../header-meta.js';
 import { fromByteBuffer } from '../header-meta.js';
-import { HttpReader } from '../http-reader.js';
+import { HttpRangeClient, HttpReader } from '../http-reader.js';
 import { calcTreeSize, type Rect } from '../packedrtree.js';
 import { buildFeature, type IFeature, type IProperties } from './feature.js';
 import { parseGeometry } from './geometry.js';
 import { inferGeometryType } from './header.js';
-import { HttpRangeClient } from '../http-reader.js';
 
 export type FromFeatureFn = (id: number, feature: Feature, header: HeaderMeta) => IFeature;
 type ReadFn = (size: number, purpose: string) => Promise<ArrayBuffer | Uint8Array>;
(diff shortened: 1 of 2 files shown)


bjornharrtell:
@stefanocudini I've fixed the root cause of https://github.com/flatgeobuf/flatgeobuf/issues/472 so if you rebase this I hope you'll get better results.

src/ts/generic/featurecollection.ts
```diff
@@ -184,6 +185,30 @@ export function buildHeader(header: HeaderMeta, crsCode = 0): Uint8Array {
...
+    const assumedHeaderLength = 2024;
+    const httpClient = new HttpRangeClient(url, nocache, headers);
+
+    const bytes = new Uint8Array(await httpClient.getRange(0, assumedHeaderLength, 'read metadata'));
+
+    if (!bytes.subarray(0, 3).every((v, i) => magicbytes[i] === v)) throw new Error('Not a FlatGeobuf file');
+
+    const bb = new flatbuffers.ByteBuffer(bytes);
+    const headerLength = bb.readUint32(magicbytes.length);
+    bb.setPosition(magicbytes.length + SIZE_PREFIX_LEN);
```
bjornharrtell: The problem is here, because of changed/fixed logic in fromByteBuffer, the position move should not include SIZE_PREFIX_LEN.

src/ts/generic/featurecollection.ts
```diff
@@ -181,9 +181,25 @@ export function buildHeader(header: HeaderMeta, crsCode = 0): Uint8Array {
     Header.addName(builder, nameOffset);
     const offset = Header.endHeader(builder);
     builder.finishSizePrefixed(offset);
+
     return builder.asUint8Array() as Uint8Array;
 }
 
+export async function readMetadata(url: string, nocache = false, headers: HeadersInit = {}): Promise<HeaderMeta> {
+    const assumedHeaderLength = 2024;
```
bjornharrtell: This assumption will not always work to get the complete header, it depends on how large schema (number of columns and names) are described by the header and other contents of the header metadata.

The true header length should be parsed and if larger than the assumed length it needs to make an additional request to get the rest.

bjornharrtell: I don't think I have test data at hand to excercise it and make the correct fix with test case... so it's not entirely trivial to fix. I might be willing to accept this PR anyway with the known caveat that it could fail on some data. @stefanocudini what is your opinion?

PR #486, opened 2026-02-08
Add RDF Ontology

reesericci:
This pull request adds an OWL (RDF) ontology for describing Flatgeobuf datasets and schemas. This allows for FGB datasets to be described as linked data, and for rich specifications to be built on top of FGB. 

For instance, we are making a universal schema for voter files on top of FGB by describing it as RDF, and having an upstream FGB onotology allows for this to be reused and shared as a common definition across datasets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ I recommend these blog posts by Horace Williams provides more details and explan
 * https://worace.works/2022/02/23/kicking-the-tires-flatgeobuf/
 * https://worace.works/2022/03/12/flatgeobuf-implementers-guide/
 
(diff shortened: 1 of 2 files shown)


reesericci:
Going to flag here that since flatgeobuf.org doesn't support content negotiations to provide the RDF, it won't be able to be consumed automatically by consumers who aren't familiar with the specification.

bjornharrtell:
@reesericci I'm sorry but I don't know much about this and don't understand the purpose/benefit. It's there some way I can learn without spending a lot of   time on it?

Issue #490, opened 2026-03-17
[ts] Create a deserialize class to have more flexibility and less parameters in functions


ger-benjamin:
Following my previous work and some discussion on Discord, I would like to introduce a ol/Deserialize class to have more flexibility in the code. Such a class could be extended, override, factorized, etc.

Also, we have sometime a lot of parameters in functions, like here: https://github.com/flatgeobuf/flatgeobuf/blob/a25978d169250a7da3e8977d24d952fdd65a8e9b/src/ts/ol.ts#L38-L47

To be clear, introduce such a class would be a breaking change for the ts part.

I'll propose a PR to have a base for discussions.

bjornharrtell:
Implemented by https://github.com/flatgeobuf/flatgeobuf/pull/496

PR #491, opened 2026-03-17
ts - Add a new Deserializer class

ger-benjamin:
Proposition for #490 

Here's my proposition to introduce a new `Deserializer` class, and related options.
In my opinion, this is great to add at this point, because this would provide more flexibility to the code.

1. This is a breaking change.
   - But the only change for the users is probably to provide an instance of the `Deserializer` and remove some other parameters.
1. We could also add a `featureProjection` option to the  `Deserializer`  class to not provide every time the wanted projection (search for `featureProjection` in this PR). But this leads to have to set the projection first, and to not rely anymore on the dynamic projection given from `ol`, in the `createLoader` method.
   - I've not used this option in my  code to not make too much changes, but I think this would be a nice to have.
1. I've not adapted the tests yet, but the vite examples are working well.

In the vite examples, it makes not a lot of differences, but as soon as you work with options in a real cases, it, IMO, really helps to have less code and a better flexibility.

So, what are you thinking about it ?

### DIFF
--- a/src/ts/deserializer.ts
+++ b/src/ts/deserializer.ts
@@ -1,0 +1,88 @@
+import type { HeaderMetaFn } from "./generic";
+import type { IFeature } from "./generic/feature";
+import {
+    type FromFeatureFn,
+    deserialize as genericDeserialize,
+    deserializeFiltered as genericDeserializeFiltered,
+    deserializeStream as genericDeserializeStream,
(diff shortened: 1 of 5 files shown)


bjornharrtell:
Replaced by https://github.com/flatgeobuf/flatgeobuf/pull/496

PR #495, opened 2026-03-31
ts - Add a new DeserializerOptions interface

ger-benjamin:
For #490 

Variante of #491 where, instead of introduce a new class, we only group parameters used by the deserialization to a single "DeserializeOptions" option and new interface.


bjornharrtell:
Replaced by https://github.com/flatgeobuf/flatgeobuf/pull/496

PR #496, opened 2026-03-31
[TS] Reworked API for deserialize

bjornharrtell:
ref https://github.com/flatgeobuf/flatgeobuf/issues/490


ger-benjamin (commented):
Looks like a good solution to me, some small comments.
I like to have the `fromFeature` as option as well.

Thank you very much.

bjornharrtell:
This will need a follow up release that fixes the examples etc.

src/ts/generic/featurecollection.ts
```diff
@@ -44,19 +44,15 @@ export function serialize(features: IFeature[]): Uint8Array {
...
-): AsyncGenerator<IFeature> {
+export async function* deserialize(input: Uint8Array, ctx: DeserializeContext): AsyncGenerator<IFeature> {
+    const { fromFeature, rect, headerMetaFn } = ctx;
+    const bytes = input;
     if (!bytes.subarray(0, 3).every((v, i) => magicbytes[i] === v)) throw new Error('Not a FlatGeobuf file');
 
     if (rect) {
         const reader = ArrayReader.open(bytes);
-        for await (const feature of reader.selectBbox(rect)) {
+        for await (const feature of reader.selectBbox(rect))
```
ger-benjamin, line 54: @bjornharrtell I think that's working, but that's strange to have not an inline `for` OR a `for` with braket here. I'm surprised that biom's linter accept this.

ger-benjamin: There are two other occurrences, but only in the `example` folder.
This is the only one I found in `src`

bjornharrtell: I don't think it's strange. I have a disposition towards not using { } for single statements, might be a bit perverse but what can I do. :D

ger-benjamin: Ok, if it's managed, it's fine to me. Thank you :-)

PR #500, opened 2026-05-07
Fix alignment issue for size-prefixed chunks

vuchl:
Adjust builder alignment to ensure size-prefixed chunk's total length is a multiple of 8 for compatibility with JS readers.

When working in a mixed environment with flatgeobuf@3.28.0 in the frontend and flatgeobuf@3.26.0 in a ASP.NET backend the frontend was unable to deserialize the flatgeobuf data.

Co-authored by Claude


bjornharrtell (approved):
Good catch, thanks.

PR #507, opened 2026-07-25
Harden Rust R-tree read path against untrusted features_count prealloc DoS

binggao1230:
The header buffer is capped (`HEADER_MAX_BUFFER_SIZE` = 10 MiB), but the Rust spatial-index read path never validated `features_count` (u64) before building the packed R-tree from it. `select_bbox_seq` / `select_bbox` / `process_index` pass `header.features_count() as usize` straight into `PackedRTree::from_buf` / `stream_search`, which:

- panicked with "Number of items too large" in `generate_level_bounds` for counts near `usize::MAX`,
- panicked with "capacity overflow" at `Vec::with_capacity(num_nodes)` for counts whose node total overflows `usize`,
- reserved `num_nodes * 40` bytes (multi-GB) up front from a 4-byte size-prefixed lie, before the first index byte was read.

A tiny malformed `.fgb` (valid magic + a header claiming a huge `features_count` with `index_node_size > 0`) makes any bbox query — the crate's headline feature — panic or attempt a huge allocation. The sequential-read fuzz target (`src/rust/fuzz/fuzz_targets/read.rs`) skips the index bytes (`io::copy(reader.take(index_size))`) and never parses the R-tree, so the index path is unfuzzed.

This is the Rust sibling of #456 (`[java] Fix int overflow issue in tree calc size`) and #505 (C++ property-bounds validation) — the Java/C++ implementations got the untrusted-input hardening; the Rust R-tree read path missed it.

## Fix (whole audited surface)

`src/rust/src/packed_r_tree.rs`:
- `validate_num_items()` rejects `num_items == 0` (`Error::NoIndex`) and `num_items > usize::MAX / 2` (`Error::IllegalHeaderSize`) at the untrusted boundary (`from_buf`, `stream_search`, `http_stream_search`), so the level-bound arithmetic can no longer overflow on a header-controlled count.
- `generate_level_bounds` asserts → `debug_assert!` — the contract holds by construction at the public boundary; trusted writer callers (`build`) still get the debug check.
- `index_size` keeps its infallible signature but is now total: returns `0` for 0 items and saturates instead of overflowing on huge counts, so the `file_reader::index_size` / `http_reader::select_all` paths can't panic either.
- `read_node_vec` takes an explicit count and caps its reservation to `MAX_INDEX_PREALLOC_BYTES` (64 MiB). `from_buf` reserves `min(num_nodes, budget)` and pushes the real `num_nodes` items off the actual bytes, so an honest index larger than the budget still round-trips while a malicious count can only make the short read fail with `UnexpectedEof`. Mirrors the `MAX_PREALLOC_BYTES` treatment in the geozero SHP/WKB readers.

`src/rust/src/file_reader.rs`:
- `select_all` / `select_all_seq` validate the count too, so the index skip can't saturate the `seek` offset.
- `read_feature` bounds `feature_size` against a new `FEATURE_MAX_BUFFER_SIZE` (10 MiB, matching the header cap) — a 4-byte size prefix can no longer trigger a ~4 GiB zero-fill resize.

`src/rust/src/http_reader/mod.rs`: `select_all` validates via `validate_num_items` before sizing the skip.

Reusing the existing `Error` variants (`NoIndex`, `IllegalHeaderSize`, `IO`) keeps the change inside the published public surface — no semver-breaking variant is added, so `cargo-semver-checks` stays green.

## Tests

`packed_r_tree.rs`:
- `from_buf_rejects_untrusted_features_count` — `usize::MAX`, `1e18`, `5e8`, and `0` each return `Err` instead of panicking (the three repro cases from the report).
- `stream_search_rejects_untrusted_features_count` — same for the seekable stream path.
- `index_size_is_total` — zero / huge / `node_size < 2` inputs no longer panic.
- `from_buf_loads_honest_index_above_prealloc_budget` — a 2M-leaf honest index (~160 MB, above the 64 MiB reserve budget) round-trips through `from_buf` with an exact extent match, guarding against regressions that would reject legitimate large indices.

`tests/read.rs`: `select_all` / `select_bbox_seq` / `select_bbox` (seekable) each reject a header claiming `features_count = u64::MAX` by returning `Err` rather than panicking.

`cargo test --workspace`, `cargo test --no-default-features`, `cargo check --target wasm32-unknown-unknown`, and `cargo fmt --check` all pass; clippy is clean.

### DIFF
--- a/src/rust/src/file_reader.rs
+++ b/src/rust/src/file_reader.rs
@@ -2,7 +2,7 @@ use crate::feature_generated::*;
 use crate::header_generated::*;
 use crate::packed_r_tree::{self, PackedRTree};
 use crate::properties_reader::FgbFeature;
-use crate::{check_magic_bytes, HEADER_MAX_BUFFER_SIZE};
+use crate::{check_magic_bytes, FEATURE_MAX_BUFFER_SIZE, HEADER_MAX_BUFFER_SIZE};
 use crate::{Error, Result};
 use fallible_streaming_iterator::FallibleStreamingIterator;
 use std::io::{self, Read, Seek, SeekFrom};
@@ -107,6 +107,7 @@ impl<R: Read> FgbReader<R> {
     ///
     /// This can be used to read from an input stream.
     pub fn select_all_seq(mut self) -> Result<FeatureIter<R, NotSeekable>> {
+        self.check_index_features_count()?;
         // skip index
         let index_size = self.index_size();
         io::copy(&mut (&mut self.reader).take(index_size), &mut io::sink())?;
@@ -129,6 +130,7 @@ impl<R: Read> FgbReader<R> {
         if header.index_node_size() == 0 || header.features_count() == 0 {
             return Err(Error::NoIndex);
         }
+        PackedRTree::validate_num_items(header.features_count() as usize)?;
         let index = PackedRTree::from_buf(
             &mut self.reader,
             header.features_count() as usize,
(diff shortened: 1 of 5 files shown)


bjornharrtell (approved):

bjornharrtell:
Looks sane to me thanks @gaoflow

PR #508, opened 2026-07-27
fixup www examples

michaelkirk:
Someone... somewhere...  (I can't find the discussion now!) mentioned that the OpenLayers example didn't update when panning. I've fixed that and also did a couple other things to polish the examples.

I did use an llm assistant, but I have reviewed all the changes.

### DIFF
--- a/examples/leaflet/filtered.html
+++ b/examples/leaflet/filtered.html
@@ -5,7 +5,7 @@
     <link rel="stylesheet" href="/examples/site.css" />
     <script src="https://unpkg.com/underscore@1.13.1/underscore-min.js"></script>
     <script src="https://unpkg.com/leaflet@1.9.4/dist/leaflet.js"></script>
-    <script src="https://unpkg.com/flatgeobuf@4.3.3/dist/flatgeobuf-geojson.min.js"></script>
+    <script src="https://unpkg.com/flatgeobuf@4.4.0/dist/flatgeobuf-geojson.min.js"></script>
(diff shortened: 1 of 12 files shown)


examples/leaflet/filtered.html
```diff
@@ -21,7 +21,7 @@
             <a href="/examples/openlayers/filtered.html">OpenLayers Example</a>
         </li>
         <li>
-            <a href="/examples/maplibre/">MapLibre Example</a>
+            <a href="/examples/maplibre/filtered.html">MapLibre Example</a>
```
michaelkirk, line 24: When viewing an example (basic, filtered, large), upon clicking a different client, we stay on that example with the new client. 

e.g. When on Leaflet/filtered, if you click on MapLibre, you'll be taken to MapLibre/filtered, not Maplibre/basic. 

I found myself wanting to navigate the docs this way, and figured most other people would too.

examples/leaflet/large.html
```diff
@@ -94,7 +94,7 @@
                         d > 50   ? '#FD8D3C' :
                         d > 25  ? '#FEB24C' :
                         d > 10   ? '#FED976' :
-                        '#FFEDA0'
+                        '#FF0000'
```
michaelkirk, line 97: Matched the color scheme across examples. It's a little discordant when flipping between them and the numbers match but the colors don't. 

I considered pulling this into a shared include, but in line with the old wisdome: "A little  copying is better than a little dependency", I think it's nicer for examples to be more stand-alone.

bjornharrtell (approved):

bjornharrtell:
Seems nice to me, thanks @michaelkirk!

PR #512, opened 2026-07-30
Fix C++ packed R-tree endian handling

Komzpa:
The standalone C++ implementation hardcodes non-GDAL builds as little-endian:

```cpp
#ifndef GDAL_COMPILATION
#define CPL_IS_LSB 1
#endif
```

That leaves the existing byte-swap paths inactive on big-endian hosts unless the file is compiled inside GDAL.

This switches the C++ implementation to FlatBuffers' own `FLATBUFFERS_LITTLEENDIAN` / `EndianScalar` helpers, which are already available through `packedrtree.h`. It also avoids mutating `_nodeItems` in-place during `streamWrite()` on big-endian builds by writing from a converted copy instead.

This was found while carrying the downstream PostGIS vendored FlatGeobuf copy on big-endian CI.

Validation:
- `cd src/cpp && clang++ -std=c++14 -Wall -Wfloat-conversion -Wmissing-declarations -fsanitize=float-divide-by-zero -g -I./ -I./include -c packedrtree.cpp -o build/packedrtree-little.o`
- `cd src/cpp && clang++ -std=c++14 -Wall -Wfloat-conversion -Wmissing-declarations -fsanitize=float-divide-by-zero -g -I./ -I./include -UFLATBUFFERS_LITTLEENDIAN -DFLATBUFFERS_LITTLEENDIAN=0 -c packedrtree.cpp -o build/packedrtree-forced-big.o`

I also tried the full C++ test target after `script/getcppdeps.sh`, but it currently fails before reaching these changes because the fetched RapidJSON 1.1.0 headers do not compile with the host compiler (`GenericStringRef::length` is const but assigned in `operator=`).

### DIFF
--- a/src/cpp/packedrtree.cpp
+++ b/src/cpp/packedrtree.cpp
@@ -29,12 +29,6 @@
 // NOTE: The upstream of this file is in
 // https://github.com/bjornharrtell/flatgeobuf/tree/master/src/cpp
 
-#ifdef GDAL_COMPILATION
-#include "cpl_port.h"
-#else
-#define CPL_IS_LSB 1
-#endif
-
 #include "packedrtree.h"
 
 #include <algorithm>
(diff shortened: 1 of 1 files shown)


bjornharrtell:
Thanks @Komzpa!

I think I've been able to fix the compile issue for the test cases in https://github.com/flatgeobuf/flatgeobuf/pull/513.

PR #303, opened 2023-09-14
Initial Go implementation of FlatGeoBuf.

bruno-ga:
- Includes a high-level API to create flatgeobuf files.
- Also a high-level API for reading those files (for now it only supports searching).
- Easily allows creating a flatgeobuf file for any source of geo-located data. All one need is a FeatureGenerator implementation.
- What is implemented seems to work but higher test coverage would be great.
- Data streaming is not implemented yet but can easily be done.

PR #516, opened 2026-08-18
fix(go): accept flatgeobuf files with a differing magic-byte patch version

### LINKED ISSUE #480: Go bindings incorrectly say "not a flatgeobuf file: invalid magic bytes"
Thank you very much for your work on this library

I am trying to read in flatgeobuf in golang. I have a file that works fine with gdal, duckdb, geopandas, etc. 

However, the go bindings says it is invalid. I am unclear why this is the case.

## Context

```
error when opening flatgeobuf error setting up flatgeobuf: not a flatgeobuf file: invalid magic bytes
```

The code that is erroring is this

```go
	fgb, err := flatgeobuf.New(filePath)
	if err != nil {
		return MainstemQueryResponse{}, fmt.Errorf("error when opening flatgeobuf %v", err)
	}

```

This is my go.mod for this library

```
	github.com/flatgeobuf/flatgeobuf/src/go v0.0.0-20251216123811-3cddc758b1df
```

If I run it through ogrinfo it prints metadata as expected.

The file that is failing is being generated from this Python code

```py
url = "https://storage.googleapis.com/national-hydrologic-geospatial-fabric-reference-hydrofabric/reference_catchments_and_flowlines.fgb"
gdf = gpd.read_file(
    url,
    engine="fiona",
    bbox=[
        -108,
        37,
        -107.5,
        37.4,
    ],
)

print(gdf)

gdf.to_file("colorado_subset.fgb", driver="FlatGeobuf")
```

youdie006:
Fixes #480

## Problem
The Go bindings reject valid FlatGeobuf files written by GDAL/geopandas with "not a flatgeobuf file: invalid magic bytes".

## Root cause
`(*FlatGeoBuf).setup()` in `src/go/flatgeobuf.go` compares the full 8-byte magic sequence against `writer.MagicBytes` (`{0x66,0x67,0x62,0x03,0x66,0x67,0x62,0x00}`). The last byte is the spec *patch* version. GDAL and geopandas write it as `0x01`, so the strict 8-byte `bytes.Equal` fails for every file they produce.

The other implementations deliberately ignore the patch byte: Rust's `check_magic_bytes` checks bytes `[0..3]`, `[4..7]` and `[3] <= VERSION`, and Java checks only the first four bytes. This leniency was the explicit, repo-wide decision in merged PR #146; the Go bindings were added later (PR #303) and missed it. Go was the only strict reader.

## Fix
Mirror the other implementations: match `"fgb"` (bytes 0-2), the major version (byte 3), and `"fgb"` (bytes 4-6), and ignore the patch version byte (byte 7). The error string is unchanged.

## Testing
Added `TestFlatGeoBuf_LenientPatchVersion`, which reads the existing `poly_landmarks.fgb` fixture, flips its patch byte to `0x01`, writes it to a temp file, and asserts `New(path)` succeeds.

- Before the fix: the new test fails with the "invalid magic bytes" error.
- After the fix: it passes.
- `go test ./...` (full src/go), `go vet ./...`, and `gofmt` are all clean.

---
This change was prepared with AI assistance and reviewed by the author.

### DIFF
--- a/src/go/flatgeobuf.go
+++ b/src/go/flatgeobuf.go
@@ -176,9 +176,16 @@ func (fgb *FlatGeoBuf) mmapOrLoadFile(path string, behavior Behavior) error {
 }
 
 func (fgb *FlatGeoBuf) setup(behavior Behavior) error {
-	// Check magic bytes.
-	if len(fgb.data) < len(writer.MagicBytes) ||
-		!bytes.Equal(fgb.data[:len(writer.MagicBytes)], writer.MagicBytes) {
+	// Check magic bytes. The last byte is the spec patch version, which the
+	// other implementations (Rust, Java) and the shared spec decision in PR
+	// #146 deliberately ignore. GDAL/geopandas write it as 0x01, so comparing
+	// the full 8 bytes would reject valid files. Match only "fgb" (bytes 0-2),
+	// the major version (byte 3), and "fgb" (bytes 4-6).
+	mb := writer.MagicBytes
+	if len(fgb.data) < len(mb) ||
+		!bytes.Equal(fgb.data[0:3], mb[0:3]) || // "fgb"
+		fgb.data[3] != mb[3] || // major version
+		!bytes.Equal(fgb.data[4:7], mb[4:7]) { // "fgb" (byte 7 patch version ignored)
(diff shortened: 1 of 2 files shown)


bjornharrtell:
Thanks @youdie006

